Issue #411, opened 2015-03-16
Tomcat 7.0.54  Psi probe not listing global datasource configured on server.xml


GoogleCodeExporter:
```
What steps will reproduce the problem?
1. Download apache-tomcat-7.0.54.tar.gz
2. tar xvf apache-tomcat-7.0.54.tar.gz
3. Change tomcat-users.xml and set manager-gui role on tomcat
5. Place latest mysql connector (mysql-connector-java-5.1.30.jar) inside lib
6. Install psi-probe
7. configure datasource on server.xml by adding inside the 
<GlobalNamingResources> tag 
the configuration

<Resource name="jdbc/test" 
auth="Container"                            
type="javax.sql.DataSource"             
maxActive="4" 
minActive="1" 
maxIdle="2" 
maxWait="10000"
username="username" 
password="password" 
driverClassName="com.mysql.jdbc.Driver"              
url="jdbc:mysql://localhost:3306/test"   
validationQuery="SELECT 1"
testOnBorrow="true" />

What is the expected result?
When I login to probe and go to http://localhost:8080/probe/datasources.htm
I should see the jdbc/test as a global configured datasource


What happens instead?
However the response is 
"There are no data sources configured for this Tomcat instance "

What version of Probe are you using?
2.3.3.

What environment (browser version, Tomcat version, JVM version, server OS)?
Java version Java 1.7.0_51, Tomcat 7.0.54 


Please provide any additional information below.

Considering Tomcat 7.0.54 

If add a resource link on the conf/context.xml file

<ResourceLink type="javax.sql.DataSource"
 name="jdbc/test" global="jdbc/test"/>

Then on the datasources page I see a listing with each application linking to 
the jdbc/test datasource defined on the server.xml (Auth Linking).

Also defining the datasource on the conf/context.xml file is successful too.
I add the same configuration on the conf/context.xml and on the datasources 
page I see a listing with each application having it's own jdbc/test datasource.
 
The exceptions that i get

02 Jun 2014 22:29:04 [http-bio-8000-exec-5] ERROR 
com.googlecode.psiprobe.beans.ResourceResolverBean  - Failed to lookup: 
UserDatabase
javax.naming.NameNotFoundException: Name [UserDatabase] is not bound in this 
Context. Unable to find [UserDatabase].
	at org.apache.naming.NamingContext.lookup(NamingContext.java:819)
	at org.apache.naming.NamingContext.lookup(NamingContext.java:167)
[...]

Issue #394, opened 2015-03-16
Support for Tomcat 8


GoogleCodeExporter:
```
What steps will reproduce the problem?
1. I copied the probe.war file into my tomcat apache-tomcat-8.0.0-RC10/webapps 
folder
2. restarted the server
3. Error : No container found for your server

What is the expected result?
Display the page without any error.

What happens instead?
getting below error message, and not able to view the process.

What version of Probe are you using?
PSI Probe 2.3.3 

What environment (browser version, Tomcat version, JVM version, server OS)?
Chrome, IE; apache-tomcat-8.0.0-RC10, Windows

Please provide any additional information below.
(Attach logs or stack traces as files instead of pasting the contents
here.)
org.springframework.web.util.NestedServletException: Request processing failed; 
nested exception is java.lang.RuntimeException: No container found for your 
server: Apache Tomcat/8.0.0-RC10
	at org.springframework.web.servlet.FrameworkServlet.processRequest(FrameworkServlet.java:583)
	at org.springframework.web.servlet.FrameworkServlet.doGet(FrameworkServlet.java:501)
	at javax.servlet.http.HttpServlet.service(HttpServlet.java:621)
	at javax.servlet.http.HttpServlet.service(HttpServlet.java:728)
	at org.apache.catalina.core.ApplicationFilterChain.internalDoFilter(ApplicationFilterChain.java:303)
	at org.apache.catalina.core.ApplicationFilterChain.doFilter(ApplicationFilterChain.java:208)
	at org.apache.tomcat.websocket.server.WsFilter.doFilter(WsFilter.java:52)
	at org.apache.catalina.core.ApplicationFilterChain.internalDoFilter(ApplicationFilterChain.java:241)
	at org.apache.catalina.core.ApplicationFilterChain.doFilter(ApplicationFilterChain.java:208)
	at com.opensymphony.sitemesh.webapp.SiteMeshFilter.doFilter(SiteMeshFilter.java:59)
	at org.apache.catalina.core.ApplicationFilterChain.internalDoFilter(ApplicationFilterChain.java:241)
	at org.apache.catalina.core.ApplicationFilterChain.doFilter(ApplicationFilterChain.java:208)
	at org.apache.catalina.core.ApplicationDispatcher.invoke(ApplicationDispatcher.java:729)
	at org.apache.catalina.core.ApplicationDispatcher.processRequest(ApplicationDispatcher.java:467)
[...]

PR #459, opened 2015-04-01
Now global resources is looked up with java:/comp/env

andresol:
I don't think this will fix (#394).


MALfunction84:
But this does fix #411?

MALfunction84:
@andresol Does this fix #411? You expressed some doubt in the issue comments. Have you tested this yourself?

MALfunction84:
Merging #464 caused this to conflict. This change might not be necessary any more.

PR #462, opened 2015-04-10
Changes to work with Eclipse

### LINKED ISSUE #16: Provide instructions for setting up the project in Eclipse
```
We should provide instructions on the wiki for how to set up the project in
Eclipse.
```

Original issue reported on code.google.com by `MALfunct...@gmail.com` on 22 Apr 2009 at 3:55

diogosantana:
These changes makes the project work with Eclipse Luna WTP to fix #16.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,7 @@
 /tomcat60adaptor/target/
 /tomcat80adaptor/target/
 /web/target/
+/.project
+/**/**.project
+/**/.classpath
+/**/.settings/
--- a/pom.xml
+++ b/pom.xml
@@ -115,6 +115,11 @@
 			</dependency>
 			<dependency>
 				<groupId>javax.servlet</groupId>
+				<artifactId>jsp-api</artifactId>
+				<version>2.0</version>
+			</dependency>
+			<dependency>
+				<groupId>javax.servlet</groupId>
 				<artifactId>jstl</artifactId>
 				<version>1.1.2</version>
 			</dependency>
--- a/web/pom.xml
+++ b/web/pom.xml
@@ -51,6 +51,16 @@
 			<type>jar</type>
 		</dependency>
 		<dependency>
+			<groupId>javax.servlet</groupId>
+			<artifactId>servlet-api</artifactId>
+			<type>jar</type>
+		</dependency>
+		<dependency>
+			<groupId>javax.servlet</groupId>
+			<artifactId>jsp-api</artifactId>
+			<type>jar</type>
+		</dependency>
+		<dependency>
 			<groupId>taglibs</groupId>
 			<artifactId>standard</artifactId>
 			<type>jar</type>
--- a/web/src/main/webapp/WEB-INF/spring-probe-security.xml
+++ b/web/src/main/webapp/WEB-INF/spring-probe-security.xml
@@ -3,7 +3,7 @@
 	   xmlns:sec="http://www.springframework.org/schema/security"
 	   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
 	   xsi:schemaLocation="http://www.springframework.org/schema/beans    http://www.springframework.org/schema/beans/spring-beans-2.5.xsd
-		                    http://www.springframework.org/schema/security http://www.springframework.org/schema/security/spring-security-2.0.6.xsd">
+		                    http://www.springframework.org/schema/security http://grepcode.com/file_/repo1.maven.org/maven2/org.springframework.security/spring-security-core/2.0.6.RELEASE/org/springframework/security/config/spring-security-2.0.6.xsd/?v=source">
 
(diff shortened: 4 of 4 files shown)


hazendaz:
Spring does have issues if you mess with the XML.  Making it validate in eclipse doesn't make it work when running it.  Those are two separate things.  There are plenty of blogs around discussing issues caused by attempting to make eclipse happy.  It doesn't have to exist to be part of the JAR in the first place.  It's best to leave it as is.  If you really look around majority of the XMLs no longer validate properly against sites.  Look at sun as a good example.  Those no longer exist but they are prevelent and we are talking about spring 2 here which is a security nightmare at this point.  I'm on task to get this on three quickly if @MALfunction84 lets me :)

.gitignores can be significantly cleaner.  It doens't need to repeat things.  Yes it needs eclipse items, intellij, and netbeans for good measure...but all the targets can be handled in on item.

This will cover the entire gitignores based on just current usage and adding eclipse.

.settings
.classpath
.project
*.log
target/

I'm just starting to build this out after a few years...glad it is on github and time to take it forwards.  Running into a number of issues already...I would say though if any of these dependencies are needed which I don't see yet due to missing dependencies...I would suggest making them all provided scope.  Anytime servlet and jsp are added to a pom, they should be provided in majority of cases.  The containers already have those and it violates spec to include them if in fact that is what ultimately happens, haven't gotten that far yet.  Just trying to get involved.  Love this program :)

diogosantana:
@hazendaz Springframework 3 and 4 schemas url are, and most of 2 are too. Eclipse doesn't complain about any other one. By the way, if you gonna upgrade to Srpingframwork, have you consider version 4?

I agree with your provided scope argument. Keep in mind that my proposal was to stop getting errors on Eclipse. Here it work well. Did you try to the project on Eclipse? Does it work?

hazendaz:
Yes, I am considering spring 4.  First going to spring 3.2.13.  I've got it up and running right now.  I've got it on spring security 3.2.7 as well but I'm not 100% certain the configuration is 100% correct so I'll need someone with spring experience to help me out there.  I'm still just learning spring but have been through a number of upgrades in the last few months to spring 3 and 4.  The only thing that has held back spring 4 has been usage of ibatis which isn't an issue here.  If you want to check that out, look over on https://github.com/hazendaz/psi-probe/tree/spring3

I've got everything working in eclipse and didn't need those libraries.  I've also reworked the build so it no longer requires using the repo's to get some dependencies.  Those are included in the build directly.  I work behind a nexus repo so unless I add those which is fine at home but not at work, it was more complicated to rely on those and since the dependencies won't change having it integrated for both travis ci (well any unix) and windows makes it an easy win.

Let me know what you think of what I got out there so far.  I plan to do a PR sometime tomorrow.  I'm about to head out for the night.

Thanks,
Jeremy

diogosantana:
@hazendaz You did a good cleanup/organization job and I agree to your changes. But I'm still getting "The superclass "javax.servlet.http.HttpServlet" was not found on the Java Build Path" jsps errors on Eclipse.

Did you migrate 100% tested to Spring 3 on ce8d4e1e40fa94932db01f3331a7127e1c15caaa?

hazendaz:
I totally missed the JSP errors.  It's been so long since I cared about JPS and that hidden gem of having java code there.  I'll add this same thing and push up in a few minutes.

I added intellij to .gitignores.

I updated maven XSD's so they are not using legacy redirects.

I squashed the 2 head spring updates.  I'm not 100% certain I got spring security completely correct but the project is working.  I'm still rather a novis with spring in many areas but at least can get around.  If you see anything there that isn't right, please let me know.  Overall it is entirely on spring 3 and using spring 3 xsds.

hazendaz:
@jander99 - The diff node issue is normal.  It sucks but that is what generally happens.  I applied the same on my PR and it is doing the same there.  I didn't cherry pick from here so at least you can see that is normal.  I've been in git for a number of years now so that is to be expected.  When you look at my PR you will also see a hard to deal with element when I tried to clean the history.  Git isn't always perfect but it is far better than what we all used to deal with :)

hazendaz:
@diogosantana Can you rework this PR?  It needs the eclipse piece removed and the spring piece removed.  The remainder should come into the project.  I'd change the PR to be labeled 'fix missing jars needed for jsp compilation' or something like that since that is technically what it is doing rather than fixing eclipse.  Or if it is easier just close this and resubmit the PR.  I believe that will get it in quicker ;)  I've got the git with support for intellij on a separate PR as well as a lot of minor issues on smaller PRs.  The remainder of what I did will come in a little slower in targetted PRs.

diogosantana:
To me is fine just close the PR and work with your changes. I can create a PR for the jsps if you will.

hazendaz:
I think you should add a PR for jsps so you get credit for finding and solving that issue.  If not I can do it tonight...probably 12 hours from now.  If I do it I'll mention you on the fix.

--- Original Message ---

PR #469, opened 2015-04-19
Spring3 and Build Enhancements

hazendaz:
I have done a number of things to enhance and upgrade the library to provide a better user experience.  This additionally pulls in changes off pull for fixing jsp errors in eclipse.  There is inconsistancy due to using various tomcats that have different jdk requirements which has not been addressed but is known and existed.  I've at least updated so it is targetting java 5 only.  It would be shocking if anyone was still using java 4 or even the legacy tomcat 5 that itself isn't even listed on the master grid at tomcats site - see http://tomcat.apache.org/whichversion.html.  I believe 5.5 pretty much replaced 5.0 so probably should look at removing that at least if not both and bumping up to java 6 would be preferred.  I do not have such an old jdk to test with nor does travis-ci.

Now for the jars in this project.  I sit behind a nexus repo so the repositories as-is don't work for me.  I'd have to either add them or stop using my own nexus.  That said, I bumped up jta since it should be a drop in place so that it uses maven central.  The jfreechart was in maven central so the repo wasn't necessary.  For oracle, I bumped to 11g as there seemed to be a conflict there and the older ones are not supported any longer from oracle anyway (even under maintainence as they discontinued them completely).  And finally I brought in third party directly to the build.  It supports unix and windows based building so hopefully that covers all.  I'm not sure with MAC.

I updated the build plugins, updated log4j to avoid exclusions, and updated junit.  I can perform junit test updates in another round.

I also added openjdk7 to the travis ci build and cleaned up git ignores to support the project better as well as provide support for intellij and eclipse.  I can add netbeans later but would require me to actually program there to figure out the ignores ;)

I cleared warnings with jmxri in the build output.

I've updated the build usage of tomcat 6, 7, and 8 dependencies to latest which gets them ahead of all vulnerabilities.  It has no impact on the build itself.

I've removed the eclipse and intellij plugins as they are either retired or on the way to retirement as IDE integration works better and those plugins are not the future.

I've updated maven's xsd's so they are not using the redirects.

And finally the real reason I'm hear today.  I updated spring and spring security to the latest in the 3 series for both.  The project is working but I need some thorough review of the spring security configuration to determine if it is completely correct.  There were a lot of changes as expected there and i've got a few things commented out.  I'm going to get someone with a little more spring experience to help review this week but anyone else that can, please do so.

I'm willing to break these into smaller PR requests or whatever.  It's not a big deal.  I ended up with a lot here based on seeing other PR with build related issues, and then having my own due to being behind nexus which is exactly the same as I'd have at my day job only there I wouldn't be allowed adding the other repos and would thus have to manually load the dependencies which is less than a perfect solution.  This solution works very well and I'd used a variation of this elsewhere.  So please provide some early feedback as my company currently has been blocked from using this product due to the spring 2 dependencies so I want to make sure we get this official and an official release so I can get it properly approved there.  In the meantime I will be testing using this version which now complies with security.

If this all works out well, I plan to start coming in and helping with whatever I can on this project to get it to 3.x and beyond.  Cheers!


diogosantana:
+1

MALfunction84:
Wow! Yes, please do break this into individual PRs for each issue resolved. Large refactors especially should be independent of other changes. If there are interdependencies, please make the sequencing clear.

Until I get some feedback from the larger community, I won't merge anything that drops legacy support until after 2.4.0 gets released. Currently, dropping legacy support is targeted at 3.0.0. This includes upgrades to Spring 3 or higher. I'm sorry, but you may need to wait for PSI Probe 3.0.0 before you can use it at your company. (However, I repeat that this is all still pending feedback from other users.)

Issue #474, opened 2015-04-20
Add OWASP maven dependency check plugin


MALfunction84:
A report on what dependencies have issues would be really valuable.

hazendaz:
To run the site, perform 'mvn site'.  This will general a basic maven site which isn't the prettiest.  I'll bring that up to par later with bootstrap using another plugin...

This will take a good deal of time to run the first time as it downloads the NVD database then scans it.  It updates generally weekly or when local cache is cleared.

Go to this location to see the report directly.

psi-probe/target/site/dependency-check-report.html

Do look over the noise produced.  There are lots of false positives but at the same time lots of positives.  Spring before 3.2.8 at any revision is vulnerable.  It's highly likely at least all compiled against versions of tomcat are vulnerable and the 5 series are unfixed.  Commons file upload is vulnerable.  The summary at the top is pretty good about pointing things out.  Just look in that grid for what looks like real hits with high confidence.  Feel free to read through them all if you have time.  PR will be pushed shortly and will reference back to this issue.  Thanks.

hazendaz:
Looks like at least the spring pieces used here got flagged low even though there are known issues in 2 series.  Anyways, high on commons file upload and tomcat 5 dependencies.  Also high on revisions of tomcat.  Some this framework only use during build time so most are easily fixable.

MALfunction84:
Closed by #476

PR #476, opened 2015-04-21
Add basic report with owasp dependency NVD Checks

hazendaz:
Owasp dependency check added to site page to scan library for
vulnerabilities.


hazendaz:
This applies to issue #474

PR #484, opened 2015-04-22
Remove jfreechart repo as it is available directly in maven central

hazendaz:


jander99:
Any version bump required or recommended?

hazendaz:
I just left the version alone for now.  If there is a newer java 1.4 it could be used in psi probe 2.4.0 but I didn't look that deep yet.

--- Original Message ---

PR #489, opened 2015-04-24
Fix classloader Leak during JFreeChart initialization

### LINKED ISSUE #443: Classloader Leak during JFreeChart initialization
```
1. Deploy probe.war
2. Access System - Memory Utilization (which starts JFreeChart)
3. Undeploy probe.war

What is the expected result?
A heap dump should not show probe classloader on the PermGen.


What happens instead?
The classloader of probe shows, which is stopped, does not get collected. See 
probeAppContextClassloaderLeak-01.png.


2.3.3 and 2.4.0-SNAPSHOT

Chrome 40 64 bits, Tomcat 7.0.57, Java 1.7.0_60 64 bits (Oracle), Windows 7 64 
bits 

The root cause can be viewed on probeAppContextClassloaderLeak-02.png

Add a patch solution as described at 
https://plumbr.eu/blog/hunting-down-memory-leaks-a-case-study

```

Original issue reported on code.google.com by `diogosan...@gmail.com` on 27 Jan 2015 at 5:37

Attachments:
* [probeAppContextClassloaderLeak-01.png](https://storage.googleapis.com/google-code-attachments/psi-probe/issue-443/comment-0/probeAppContextClassloaderLeak-01.png)
* [probeAppContextClassloaderLeak-02.png](https://storage.googleapis.com/google-code-attachments/psi-probe/issue-443/comment-0/probeAppContextClassloaderLeak-02.png)
* [psi-probe-patch.txt](https://storage.googleapis.com/google-code-attachments/psi-probe/issue-443/comment-0/psi-probe-patch.txt)

diogosantana:
Fixed #443.

### DIFF
--- a/core/src/main/java/com/googlecode/psiprobe/AwtAppContextClassloaderListener.java
+++ b/core/src/main/java/com/googlecode/psiprobe/AwtAppContextClassloaderListener.java
@@ -1,0 +1,51 @@
+package com.googlecode.psiprobe;
+
+import javax.imageio.ImageIO;
+import javax.servlet.ServletContextEvent;
+import javax.servlet.ServletContextListener;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+
+/**
+ * Prevents a classloader leak as suggested by <a href="https://cdivilly.wordpress.com/2012/04/23/permgen-memory-leak/">Colm Divilly</a>
+ * 
+ * @author diogosantana
+ *
+ */
+public class AwtAppContextClassloaderListener implements ServletContextListener {
+
+	
+	public void contextInitialized(ServletContextEvent sce) {
+		try {
+			final ClassLoader active = Thread.currentThread().getContextClassLoader();
+			try {
+				// Find the root classloader
+				ClassLoader root = active;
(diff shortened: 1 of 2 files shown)


hazendaz:
I realize this was pulled in but wondered if any thought had been given to including this listener?  It effectively does the same and more.

https://github.com/mjiderhamn/classloader-leak-prevention

diogosantana:
Seams to be a good framework. When I proposed the patch, I didn't hit this
framework on my research. It would save me a lot of time.



Diogo Sant'Ana

hazendaz:
I've been running this framework for a couple of years.  Works great for most setups.  It doesn't work in all cases and it mentions that but probably is safe here.  If it was to be used would need some extended testing unloading and reloading psi probe to make sure no stack traces.  It does a lot and at the moment not configurable enough in case not everything is needed.  I've seen it fail with mule esb but has worked with everything else I've tried it with.

--- Original Message ---

diogosantana:
Could you create a issue to consider the use of it?

hazendaz:
Sure.  It sounds like a great thing to try given what this library is for.  I'll create an issue later this weekend.

--- Original Message ---

Issue #482, opened 2015-04-21
Add servlet-api and jsp-api dependencies to web project


diogosantana:
Add servlet-api and jsp-api dependencies to web project.

@hazendaz I still getting the spring-security xsd not found on validation. But since .settings is on gitignore, I can suspend xml validation and get no modified file. That works for me, until Spring 3.

MALfunction84:
These should be indirect dependencies of web already.

* web => core => servlet-api-2.4
* web => core => jasper-compiler => jsp-api-2.0

What's the motivation for adding them directly to web?

diogosantana:
Both are provided in core and not visible in web.

Try to add a servlet to web a run mvn clean package. It won't compile web.

    package web;
    
    import java.io.IOException;
    
    import javax.servlet.ServletException;
    import javax.servlet.http.HttpServlet;
    import javax.servlet.http.HttpServletRequest;
    import javax.servlet.http.HttpServletResponse;
    
    public class TestServlet extends HttpServlet {
    
    	private static final long serialVersionUID = 3412500610604747990L;
    
    	protected void doPost(HttpServletRequest req, HttpServletResponse resp) throws ServletException, IOException {
    		System.out.println("teste");
    	}
    }

Compile message:

    TestServlet.java:[5,20] error: package javax.servlet does not exist

hazendaz:
Yup it is jsp related.  It has a huge amount of errors otherwise in eclipse.  This is the fix.

--- Original Message ---

MALfunction84:
> Try to add a servlet to web a run mvn clean package. It won't compile web.

There is currently no Java source in the web module. I don't understand the desire to to add a dependency that wouldn't be used. But I guess if Eclipse requires it...

hazendaz:
Jsp code.  Eclipse validates and flags as an error.  This just clears the error.  It didn't affect the maven build itself.  In jsp eclipse allows code completion like elsewhere.  That said it cannot do its job without this.  So we are all good now.  Longer term expression language should be use in jsp rather than java code...

--- Original Message ---

PR #491, opened 2015-04-24
Use JTA 1.1 plus minor project changes

hazendaz:
JTA 1.1 can be used directly from maven central and still targets java 1.4

Added openjdk7 to travis ci build
Added user group to README
Fixed formatting in POM from a prior change (excess tabs).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,5 +2,5 @@ language: java
 jdk:
   - oraclejdk8
   - oraclejdk7
+  - openjdk7
   - openjdk6
-  
--- a/README.md
+++ b/README.md
@@ -21,3 +21,7 @@
 		mvn package
 
 	This will create a deployable file at `web/target/probe.war`.
+
+# User Group
+
+Please follow the user group on google groups: [psi-probe](http://groups.google.com/group/psi-probe/)
--- a/pom.xml
+++ b/pom.xml
@@ -54,17 +54,6 @@
 	</modules>
 	<repositories>
 		<repository>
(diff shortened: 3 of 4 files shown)


diogosantana:
@hazendaz You removed servlet-api and jsp-api from web/pom.xml reverting #482.

hazendaz:
It's still there.  Formatting was off by extra tab on each line.  Just formatted to match the file.

--- Original Message ---

diogosantana:
I even have deleted my comment, because I didn't pay enough attetion and just saw the deleted lines.

MALfunction84:
Doesn't JTA 1.1 require Java 5+?

hazendaz:
It was built with a very early jdk 5 but it states it is for java 2.  So that tells me 1.4.

--- Original Message ---

hazendaz:
I've got eclipse setup to tell me if a incompatible item is used.  JTA 1.1 is good with java 1.4.  

For reference, off the master branch, this one is not reported as an issue.  To be on the safe side, I'm going to see about downloading 1.4 and 1.5 to get a better consistent outcome as moving up in jdks cuases a huge inability to fully understand if it is good or not.  I'll report back here shortly but I'm pretty positive this is jdk 1.4 compliant.

hazendaz:
I've downloaded java 1.4 and 1.5.  JTA is good.

ucp which isn't part of this is wrong.  It requires 1.5.  Tomcat 7 & 8 both require 6 and 7 respectively.

So this change in other words is 100% compatible with java 1.4.

Issue #453, opened 2015-03-19
Upgrade to Maven 3


MALfunction84:
Maven 2 is ancient. The psi-probe project should upgrade to Maven 3.

MALfunction84:
Turns out the Travis CI build works just fine.

PR #492, opened 2015-04-25
Junit 4.12

hazendaz:
Per #453, this library is using maven 3.  Therefore, it can use the latest compiler.  Additionally, this opens the door to continue compiling source/target for 1.4 java while treating test as source/target 1.5.  This allows for the library to be upgraded to junit 4.12 through a bit of maven magic.  This does not affect the outcome of the build itself and still is compliant for java 1.4.  All JVMs used to build are 1.6 and above which means this is suitable for inclusion into the build process even with psi-probe 2.4.0.

### DIFF
--- a/core/src/test/java/com/googlecode/psiprobe/beans/stats/listeners/FlapListenerTests.java
+++ b/core/src/test/java/com/googlecode/psiprobe/beans/stats/listeners/FlapListenerTests.java
@@ -10,14 +10,14 @@
  */
 package com.googlecode.psiprobe.beans.stats.listeners;
 
-import junit.framework.Assert;
-import junit.framework.TestCase;
+import org.junit.Assert;
+import org.junit.Test;
 
 /**
  *
  * @author Mark Lewis
  */
-public class FlapListenerTests extends TestCase {
+public class FlapListenerTests {
 
     private final int defaultThreshold = 10;
     private final int defaultInterval = 10;
@@ -34,13 +34,14 @@ public class FlapListenerTests extends TestCase {
         listener.reset();
         add(sce, defaultInterval);
     }
-    
+
     protected void add(StatsCollectionEvent sce, int quantity) {
         for (int i = 0; i < quantity; i++) {
             listener.statsCollected(sce);
         }
     }
-    
+
+    @Test
     public void testBelowThresholdNotFlapping() {
         listener.reset();
         listener.statsCollected(aboveThreshold);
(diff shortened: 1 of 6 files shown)


diogosantana:
After this merge I'm getting erros on Eclipse: Syntax error, annotations are only available if source level is 1.5 or greater.

This is because the use of `@Test` in test classes even though maven is configured with 1.5 for test.

Did anyone get this too? There is a know fix for this? I searched without sucess.

hazendaz:
What command are you running and what maven version?

--- Original Message ---

diogosantana:
No maven involve direct. I am using the last version of m2eclipse shipped
within Eclipse Luna.
The error is from Eclipse compiler.


Diogo Sant'Ana

hazendaz:
So the build is working or not?  Ignore syntax violations.  Depending what you have set there are plenty with this library.  If mvn clean install works you are good.  It works on Travis ci and my machine.  I've done this elsewhere too.  If it's just m2e complaints I wouldn't worry about it.  I'll get on in a few to see what I see on my end in case there is a real build issue.

--- Original Message ---

diogosantana:
I know it a m2e issue and maven package works ok. My question is about a
workaround on Eclipse. Nobody likes erros messages...


Diogo Sant'Ana

hazendaz:
I understand what you are saying.  I hate seeing errors as well.  However, I overlook those when they don’t apply to the source code itself.  I’m assuming either you have some other check picking that up or I’m not seeing it because I use Eclipse Mars which comes with newer m2e.  If it builds ok, which it does, then we are ok for now.  I just had a laundry list of things I wanted to try to get into the 2.4 release.  Next I’m going to try to tackle as many issues as possible.  And hopefully with any luck that will get a 2.4 release in the next month or so.  Especially given it’s been about 2 years.  As soon as this library hits java 5, say version 2.5, then this specific issue will go away as well as a bunch of compliancy issues with the java versions being used.  JDK 1.4 simply put is fine if the library as a whole complied there and after a lot of research it does not.  It’s a real mixed back but as of right now I have tested all the way back to java 4 and tomcat 5.0.28.  Everything is functioning correctly at this point.

 

If you can get some screen shots of what you are seeing that might be helpful.  You can email me at jeremylandis@hotmail.com <mailto:jeremylandis@hotmail.com>  as I’m not sure how easy that is to send through here and have it display properly.

 

Thanks,

 

Jeremy

PR #494, opened 2015-04-25
Added versions-maven-plugin and report functionality. 

jander99:
This plugin will allow us to see which dependencies and plugins are out of date and update them appropriately.  This will not recognize dependencies where the groupId and/or artifactId have changed in maven, e.g. javax libs.


hazendaz:
Sorry missed you had this separately.  Back on the other, I would remove that. Keeping the changes separate.

jander99:
I will close #495 and rework it.  Please commit or close this PR so I can also move forward on rewriting the reporting section of pom.xml according to [this post](http://gordondickens.com/wordpress/2012/08/02/maven-3-code-analysis-and-reporting/)

hazendaz:
Since this is rather isolated from changes on #495, you could just branch off the master, make the other changes and do a PR there.  Then on this one, just make the property change for it by itself.  That way there is no dependency between the two.  I'm pretty sure that would work out correctly.  I've only recently started really trying to use branches and it is making things much easier so I don't sit and wait or worse end up with the world on a PR.  Actually seems to work better that way if I can keep changes pretty isolated.  That is really hard when working on poms though...

jander99:
I'm going to do more than one PRs for version bumping.  First PR will be simply to place everything in the <property> tags with no version changes.  Then we can go from there to bump any old versions that need to be changed.  That way there aren't many changes per PR.  Git is new to me. Learning curve is starting to flatten out though, so that's good.

hazendaz:
I added a new issue regarding jdk 1.4 compliancy.  I’ve also pulled java 4 and java 5 so I can properly test.  I got a copy of tomcat 5 and will get 5.5 and 6 tomorrow so I can verify across all containers.  Check that last issue out I created as it deals with a couple versions you had made changes to the groupID.  In one case that raises the compliancy level required.  That isn’t to say there are not already a number of issues as there are.  I’ve got a cut off to the side with all that updated and notes where it doesn’t comply with jdk 1.4.  After I get through doing a full regression in each version, I can send a PR on the verion bumping.  I listed it all there so if you see something missing let me know.  I think I got it all straight.  Personally, not sure why we are even trying to support jdk 1.4.  Anyone using that still doesn’t much care about security and obviously can survive forever on older versions of this library.  At best, maybe we pin this down with what can still go up as 1.4, just note the known noncompliance, get product lead to release, then move on to what I feel we care more about, newer stuff.  I for one will certainly play the old numbers game but hope it is extremely short lived.  The maven team had a similar discussion recently and finally decided to just skip up to java 7 after they hit the 2.2.1 maven revision.  That went back and forth forever.  In this day and age, with GIT, we don’t need to support old copies any longer.  That is what branches are for and people who choose to drag along certainly can fork and move on their own path.  All that said, I’d love to see version 2.4 of this cut shortly so we can actually move on!  In the meantime, lots of GIT practice as well as learning this library which is helpful for all of us :)

 

Jeremy

hazendaz:
I put a  PR out there with versions bumped based on findings on figuring out last working with jdk 1.4 across the board.  You might want to wait a day or so on this until that gets merged that or just look at those and try to target it afterwards with those changes.  I think we will be pretty straight on versions for 2.4 release then.  I’m going to switch gears at this point and start trying to knock out some of the issues.  It would be really cool if we could bring that down to a manageable number of issues prior to 2.4 release.  Doing so probably will open the door for moving this library up to newer technologies…

PR #495, opened 2015-04-25
Version updates

jander99:
This PR will supercede #494 and should be applied after #491, #492, and #493.


MALfunction84:
What is gained by doing this? Autocompletion of version numbers in IDEs (mine at least) occurs within the <version> tag.

I can see the value of ensuring that e.g. Tomcat artifact versions are consistent across or within modules, but for everything else, I'm struggling to understand the motivation behind this change. @jander99 Can you elaborate?

jander99:
I've started reading up on maven, and some best practices and whatnot.  It seemed to be a best practice to pull version numbers out and place them in a variable to be referenced.  I was also playing with a versions plugin(PR #494 ) and having all the versions in one place made it easier to update.  

The versions referenced in this PR are version bumps based on the output of the version plugin.  I did not bump each artifact very much, erring on the side of caution.  #495 has the change from #494 so you can accept one or the other in order to get the report functionality (which I am redoing in maven 3 format next) or both the report functionality and the version bumps.

hazendaz:
Can you break this into clear PR requests?  I see a number of things here.

1) Setting properties for versions.  This is maven suggested best practice and I fully agree.  Let's keep this PR to just that and maybe relabel it as such.

2) Adding different group id's in some locations.  I'd say pull that out to another PR request ensuring that they still target 1.4.  Since that question came up on JTA on my PR, eclipse has a setting with luna to tell us when we are using the wrong jars.  There are a number of issues regardling that but due to tomcat versions rather than others.  So easily enough to very by setting that check to verify incompatible class files.  If all are good and the group changes are appropriate, on another PR would make that clear that it is only version changes targetting 1.4.

3) Some new plugins are being introduced.  Versions plugin which doens't need special additional configuration.  However, I'm thinking a PR that has that plus all other available reporting we can would be a huge win.  See https://github.com/hazendaz/base-parent and take all you need.  Run that guy once and look over the site to see what you may want.  It likely will be most of it if not all.  The only thing I didn't get quite working right was the pdf document but all the other elements work.

jander99:
Closing this to rework it.

Issue #395, opened 2015-03-16
Add support for TomcatSlf4jLogback


GoogleCodeExporter:
```
What steps will reproduce the problem?
1.Tomcat-7.0.47
2.tomcat-7.0.47-slf4j-1.7.5-logback-1.0.13.zip 
(https://github.com/grgrzybek/tomcat-slf4j-logback)

What is the expected result?
Log files from logback will be shown (tomcat and installed application)

What happens instead?
Only probe.log is shown. In probe.log is error shown.

What version of Probe are you using?
2.3.3

What environment (browser version, Tomcat version, JVM version, server OS)?
Any browser, 7.0.47, 1.7.0_25-b17, Windows 7 (Service Pack 1)

Please provide any additional information below.
(Attach logs or stack traces as files instead of pasting the contents
here.)

```

Original issue reported on code.google.com by `ondrej.t...@gmail.com` on 10 Jan 2014 at 2:34

Attachments:
* [probe.log](https://storage.googleapis.com/google-code-attachments/psi-probe/issue-395/comment-0/probe.log)

jander99:
Scanning Milestone-2.4.0.

Should we move this to 3.0 and work on adding supported logging frameworks?

diogosantana:
+1 to move to 3.0

PR #498, opened 2015-04-27
TomcatSlf4jLogback Support

hazendaz:
Tomcat-slf4j-logback project replaces tomcat juli with slf4j/logback. See [here](https://github.com/grgrzybek/tomcat-slf4j-logback)

This results in inability for psi-probe to see those logs.  This change
treats the logs similar to logback and thus makes them available and
accessible.


MALfunction84:
Does this fix #395?

hazendaz:
Yes it does solve that issue.  I didn't know there was a ticket already.  I am a committer on that project so wanted it for my needs.  Glad to see others are using it along with this library.

--- Original Message ---

PR #502, opened 2015-04-29
User Groups links to open in new tab

diogosantana:
This makes the User Groups links to open in new tab.


MALfunction84:
If I accept this change, there's no easy way to "opt out" of the link opening in a new window. Currently, you can middle-click the link to "opt in." I think I prefer it that way.

Issue #501, opened 2015-04-29
Update artifacts preserving compatibility with JDK 1.4


hazendaz:
I have Eclipse setup to report incompatible java versions in use.  Aside for the obvious for tomcat -> see here http://tomcat.apache.org/whichversion.html, there are many others in issue.  I will have a pull request on this later.  This is what I've learned.

junit can go to 3.8.2
jstl can go to 1.2.  This requires including javax.el version 1.1 as tomcat 5.x doesn't have this.  May have to be careful here unless tomcat will exclude.  I'll be testing this.
mail can go to 1.4.7

ucp requires java 1.5
tomcat-jdbc requires java 1.5
bonecp requires java 1.6

commons-collections can go to 3.2.1
commons-fileupload can go to 1.2.2
commons-lang can go to 2.6
commons-logging can go to 1.2 which also fixes vulnerabilities.
commons-modeler can go to 2.0.1 and loses two exclusions
jaxen can go to 1.1.6
quartz can go to 1.6.3
sitemesh can go to 2.4.2
xpp3_min can go to 1.1.4c
xstream can go to 1.4.1
c3p0 must stay under this groupId and can go to 0.9.1.2
commons-dbcp can go to 1.3 and loses its exclusion
wrapper can go to 3.2.3
jfree gets org on the groupId and can go to 1.0.15. 

tomcat60adaptor snuck in ecj 4.3.1 which now requires 1.6 so just exclude it so it continues the 1.5 support.
tomcat70adaptor should be 1.6
tomcat80adaptor should be 1.7

If tomcat versions are corrected, it would also mean we cannot build using jdk 6 but rather 7 for the build.

There are some other dependencies that end up getting pulled in as well so some research would need to be made above this to determine what can be safely excluded so additional jars are not needlessly added.

At the moment I have tested this on tomcat 7 to no ill effects.  I'll continue to test my way down.  I've started pulling older tomcats as well so I can ensure this is working properly across the supported line.

PR #503, opened 2015-04-30
Bring up dependencies to 1.4 compliancy throughout

hazendaz:
A number of minor issues resolved as well.  The primary thing here was getting this up to full 1.4 compliancy as we can get.  Further, I've added notes about pieces that are not compliant.  And further still, I have tested this on java 4 and tomcat 5 as well as up through other tomcats.  All works well.

Spring 2 xsd just move back to one spring actually has.
Spelling correction
Fix timestamp for maven build so it shows on psi-probe landing page.
Update copyright year to 2015
Fix project location from old googlecode.


hazendaz:
Fixes #501

Issue #497, opened 2015-04-26
Maven Reporting - All Modules


jander99:
Please work on having all modules build a site and aggregating data together.  

Also, during site build, would be neat to have it commit to gh-pages. Is that possible via travis?

hazendaz:
It is possible to commit to gh-pages.  Do you want the current gh-pages merged into the site build?  Travis build should be able to do that just by adding a hook for it.  Probably will not want that on every build to travis CI so maybe a special branch - ie site so it will only perform that when expected.

jander99:
I played around a little with trying to get the reporting to roll up from each module, but I guess my maven skills are still a little weak.  It would be up to @MALfunction84 as to whether he wants the site output put on gh-pages.

MALfunction84:
Unless I'm mistaken, when I perform `mvn site`, it churns for a long, long time, but it eventually builds a single site for all modules. @jander99 were you expecting something different?

@hazendaz I agree that we probably don't want the site updating on every build, but I also don't like idea of having to merge to a branch each time we do want it to update. The current gh-pages can be overwritten. It's just a copy of the wiki homepage.

jander99:
Some reports aren't aggregating correctly.  I am going to investigate.

hazendaz:
We may want to disable nvd scan on site unless we need it.  That is the reason it runs slow.  There is a lot of other reporting that is useful as well that doesn't slow it down.  Running on each build is ok.  My concern there was due to it replacing gh pages.  If not a concern then that is good enough.  May want to have it run though on merges to master rather than pull requests if that is possible.  If Travis ci does it time to complete isn't that big of a deal as far as nvd scan goes.

--- Original Message ---

hazendaz:
There is a big in maven getting multi module site generation to properly link.  So off the main site pages others are not reachable.  If you can figure out a solution with that it would be awesome...

--- Original Message ---

jander99:
We can call this closed, and if anyone wants to tweak later, they can, but #507 was supposed to address most of this.

MALfunction84:
Instead of disabling the dependency-check report entirely, couldn't we just exclude it by default and then create a profile that includes it?

jander99:
My maven-fu is not that strong yet.  I think we could have a
profile/goal/whatever that executes everything and keep package to only
those bits that are required for generating the war file.

hazendaz:
I can take care of profiling that one

--- Original Message ---

PR #507, opened 2015-05-11
Update project pom to include reporting plugins for mvn site.

jander99:
Update project pom to include reporting plugins for mvn site.

Don't ask how the extra spaces in the <project> tag got added. I can't seem to undo it in IntelliJ.  

Disabled org.owasp:dependency-check-maven as it causes builds to be long. We can think about adding it for releases or as something optional when we start updating artifacts to latest versions.


MALfunction84:
This seems to contribute some toward resolving #497.

Issue #490, opened 2015-04-24
Psi probe not listing global/application datasources - tomcat


mbonastre:
Maybe this issue is redundant; there are other open issues about datasources, but they are very specific and don't seem to fit this case.

Psi-Probe does not correctly list datasources in Tomcat 6 and 7. In tomcat 8, some datasources are listed, others are not.

The commit d4cceeabf847b449848648639c7bc22c4c38dabd tries to solve it, but it only works partially on Tomcat 8.

I have tested this on:
 - Tomcat 6.0.43
 - Tomcat 7.0.61
 - Tomcat 8.0.21
With probe-2.4.0-SNAPSHOT (20150423 08:59)

What steps will reproduce the problem?
 - Install probe on default tomcat config as described in probe wiki
 - Create one application datasource in file conf/Catalina/localhost/docs.xml

<Context>
<Resource name="jdbc/appOracle" auth="Container"
              type="javax.sql.DataSource" driverClassName="oracle.jdbc.OracleDriver"
              url="jdbc:oracle:thin:@(description=(...))"
              username="xxx" password="xxxx" maxActive="20" maxIdle="10"
              maxWait="-1"/>
</Context>

 - Create on global datasource in GlobalNamingResources in conf/server.xml

<Resource name="jdbc/globalOracle" auth="Container"
              type="javax.sql.DataSource" driverClassName="oracle.jdbc.OracleDriver"
              url="jdbc:oracle:thin:@(description=(...))"
              username="xxx" password="xxxx" maxActive="20" maxIdle="10"
              maxWait="-1"/>
</Context>

 - Add Oracle driver to lib
 - Start tomcat
 - Login to probe
 - Go to "Data Sources" tab

What is the expected result?
  - You should see two datasources: one global, one application

What happens instead?

In tomcat 6 and tomcat 8:
  - Probe shows message "There are no data sources configured for this Tomcat instance"

In tomcat 7:
 - Probe shows only the global datasource

Also, in Tomcat 6 and 7:
  - If you go to Tab "Applications", click on "/docs", select "Resources":
    Shows datasource "jdbc/appOracle" with error:
    "This resource is misconfigured and runtime information is not available"

  - If you then click on "jdbc/appOracle", "Test Connection":
    Shows error "Data source jdbc/appOracle does not exist"

In Tomcat 8:
  - If you go to Tab "Applications", click on "/docs", select "Resources":
    Seems to be ok.
[...]

PR #509, opened 2015-05-12
Optimize imports.

jander99:
Run IntelliJ IDEA "Optimize imports" tool.  This tool reorganizes the import statements and removes those that are no longer needed.


jander99:
Yeah, this PR is huge.  It's more of a test of our resolve when there are larger changesets that come through during 3.x development.

hazendaz:
Ummm...intelliJ import cleanup scares me.  I've washed it trash a library too many times and that is using 14.x.  Eclipse does this cleanup too and I've never seen it fail.  You seem to be heavy intellij user so maybe something I did wrong when playing around with it.  Otherwise no concerns from me.  If we are going the cleanup route might be a few things happening as we get there.  All will certainly be huge.

MALfunction84:
This is definitely needed, because it's not at all consistent. What are our options across IDEs for how it's configured? Besides just moving imports around, I want to get the configuration committed so people who check out the project don't have to take any extra steps. (Same for code formatting, but that's another topic.)

IntelliJ seems to have weird ideas about the order imports should come in. I noticed it puts java.* and javax.* at the bottom, then everything else at the top. Doesn't Eclipse do the reverse? Or is that configurable? I know it's configurable in NetBeans, including whether it puts empty lines between the groups. By default it just puts things in alphabetical order with no empty lines.

I don't have a strong preference about the ordering of imports or how they're grouped. Please chime in. Would it be worthwhile to have the PSI Probe imports grouped?

jander99:
I'm at work right now, so only have access to Eclipse.  IntelliJ's help gives instructions on how to export settings to a jar file:  https://www.jetbrains.com/idea/help/exporting-and-importing-settings.html
I think Eclipse exports to an xml file.  Not sure about Netbeans but probably similar.  

Sounds like someone or a group of someones will need to come up with 3 files, one for each IDE.  

There is also https://github.com/revelc/formatter-maven-plugin but it looks like it only does Java, not xml/js/css/etc.

hazendaz:
You discovered some of my work...you are right java formatter maven plugin in only java.  There is another fork which I think we are going to reach out and try to bring together.  This formatter works independent of eclipse.  So it works across all three ide's.  I'm an eclipse guy so I like it.  I think though we really only want one formatter though.  No real preference otherwise.

--- Original Message ---

jander99:
Well, how about that?

I did find this.  Thoughts?

https://code.google.com/p/google-styleguide/source/browse/trunk

hazendaz:
Junit applied google format.  I'm all for it.

--- Original Message ---

jander99:
I will see about importing this to IntelliJ and updating this PR.  I'll
post on the forum the location of the guides.  We can figure out there how
best to distribute it.

Want to keep the issue on message and the forum should be the goto play to
discuss.

MALfunction84:
If we're indeed going with Google formatting, this PR is not the [correct ordering of imports](https://google-styleguide.googlecode.com/svn/trunk/javaguide.html#s3.3.3-import-ordering-and-spacing). ```javax``` should come after ```java```.

Also, their grouping of ```com.google``` imports at the top would be analogous to us grouping ```com.googlecode.psiprobe``` at the top. (Note to self: update package names in 3.0 to something a little more future-proof.) Thoughts?

jander99:
Yeah keep this PR on ice for a few days. I'll get around to checking out
the features IntelliJ has to import a formatting XML document.  Hopefully
it's that easy.  Otherwise, I'll close the PR and we can keep discussing.

MALfunction84:
Reordering imports is one thing. [Reformatting all of the source is another](https://www.youtube.com/watch?v=V_jy9atgThc&feature=youtu.be&t=146). I mentioned in #512 that we probably shouldn't apply the reformatting until just before we create the 2.4 branch. This way it doesn't create enormous conflicts for anyone who has a fork of master and is working on 2.4's final features. Any thoughts on that timing? Probably should discuss in #512.

jander99:
You require additional pylons.

hazendaz:
I'd say we wait until after 2.4 so it's far easier to track down any bugs found in the release.  Then do that early in the next round.  Also cut a branch of 2.4 at that time just in case regressions found later.

--- Original Message ---

jander99:
I can agree with that sentiment.  I'll close this PR and we can revisit later.  

However, I would like to get the discussion going on whether or not we are good on 2.4.  I have it deployed to Tomcat 8 with no issues, however I am not using it to its fullest abilities as a replacement for manager/ and host-manager/.

MALfunction84:
@hazendaz Agreed, mostly. I'm not clear on how applying the reformat after 2.4 helps us track down bugs, though.

I was suggesting we do the reformat, then create a 2.4 branch, then tag/release 2.4.0. This way, master and the 2.4 branch have the same formatting which should reduce the friction in merging fixes between the two if that becomes necessary later. We'll still have the 2.4 branch and 2.4.0 tag to be able to make sense of any stacktraces we see.

The [2.4.0 Milestone](https://github.com/psi-probe/psi-probe/milestones/2.4.0) has the remaining issues. We can probably close #394, but #490 seems to affect several versions, including Tomcat 8.

hazendaz:
Valid points on when to do it.  At end sounds better now.

--- Original Message ---

hazendaz:
Details on the google way…

 

http://google-styleguide.googlecode.com/svn/trunk/javaguide.html

PR #514, opened 2015-05-15
Build Enhancements - All Maven

hazendaz:


hazendaz:
Opened issue #515 for this effort.  Should have opened that first and will do so on next one.

jander99:
These version bumps are the same as what I previously did and tested with, so I can approve them.  Mostly minor fixes anyway.

MALfunction84:
Looks like the tomcat80adaptor is missing the ```git.relative.path``` property.

MALfunction84:
Also, is there a way to get that ```copyright``` Maven property into [messages.properties](https://github.com/psi-probe/psi-probe/blob/master/web/src/main/conf/WEB-INF/messages.properties) (and all its translations)? The ```probe.jsp.copyright``` has a year in there that I have to update all the time.

MALfunction84:
This is phenomenal work otherwise. Apologies for the multiple comments. Once the ```git.relative.path``` issue is fixed, I'll be happy to merge. The copyright one could wait, but I'll definitely merge if you fix it too.

hazendaz:
git.relative.path - added and squashed commit.

For using copyright in probe.jsp.copyright, it is easy enough but have to ask question about that.  The web portion has both filtered and unfiltered pieces.  Do you know why?  Basically to solve the copyright issue, it is as easy as adding ${copyright} to each of those property files but at the moment the project has filtering turned off there.  So that makes me wonder if there was a reason to do so.  I have tested this out and it works just fine but wanted to understand the original filtering rules when bundling up the war.  If no real issue, maybe those should come back together.  I'm not seeing a good reason to have those separated as the syntax used doesn't otherwise exist in non filtered property files.  I certainly can pull these together on a separate PR.

MALfunction84:
Originally nothing was filtered. I pulled out the version into a filtered path to get the version number from Maven to show up in the final build. I don't see any harm in combining the two.

This all looks good. Thanks!

Issue #505, opened 2015-05-01
Resource bundles in non-standard location


jander99:
With a "standard" maven directory structure, message resource bundles typically go in /src/main/resources/ and not /src/main/conf or /src/main/conf-filtered.  The maven-resources-plugin should be able to handle filtering messages. We just need to update the application to look in /src/main/resources instead of /src/main/conf.

Please tag for 3.0.0. For now, l10n-maven-plugin as part of #504 will not pick up /src/main/conf.  This can possibly be passed via a configuration, but the better solution is to move the resource bundles to the "more proper" location.

hazendaz:
PR #517 combines /src/main/conf and /src/main/conf-filtered.  I do agree they are in wrong place but don't really see a point to waiting until 3.0.0 to fix.  So maybe after #517 is on board, we can refactor and fix this.  It should be pretty easy and is just a build item.  In fact, they can stay as is and simply move to resources.  It will copy automatically anyways to the correct location then as it will correctly match the final layout.  Filtering should work just the same.  Had I noticed this first, I think I would have just did this in one go.

hazendaz:
Original issue after testing and verification is incorrect in proper location.  That works for a JAR.  Being this is a WAR, an issue still exists but actually things just need to move under src/main/webapp.  I've done the change and verified it works in tomcat. I did try the original suggestion first with a number of modifications and it would not work.  Things just got moved around too deeply in the end and/or required far more changes than necessary.  The new build though has 1 less file.  Originally the context.xml was getting put in two places.  It stated specifically to put under META-INF for jboss but actually that is the same for jboss and tomcat.  Feel free to correct me if I missed anything and check out the PR.

hazendaz:
PR didn't work out as it is auto changing line endings and various tries failed to get that to stop occurring.  So this will be on hold until after code base is normalized.  Everything works as-is now so this is just a cleanup effort.

MALfunction84:
As far as I can tell, files in src/main/resources end up being copied into WEB-INF/classes when you're dealing with the maven-war-plugin, and no filtering is applied. Some of the resources (messages*.properties and version.properties) need to be copied to WEB-INF and filtered. Others just need to be filtered (log4j.properties). If there is a way to do these without using ```<webResource>```, I don't know of it.

[...]

Issue #83, opened 2015-03-16
Support for JBoss Application Server 3.2, 4.0, 4.2, 4.3, 5.0, 5.1


GoogleCodeExporter:
```
What steps will reproduce the problem?
1. Download Jboss 5.1.0 GA
2. deploy probe 2.1
3. Go to localhost:8080/probe/



What is the expected output?

Expect probe ui to be displayed. 

What do you see instead?
Exception in the jboss longs.


What version of the product are you using? On what operating system and
browser?


Please provide any additional information below.

```

Original issue reported on code.google.com by `neeraj...@gmail.com` on 21 Jan 2010 at 8:04
- Blocked on: #103

PR #518, opened 2015-05-21
Fix issue #505 correct locations of resources

hazendaz:
Same results except that context.xml is not duplicated under WEB-INF.  Being
that it was reported the correct location was under META-INF for jboss only and having
used context xml in that location back to tomcat 6, I believe this is
ok.  I do not know going back to tomcat 5 or 5.5 if that was true but certainly is for 6+.

btw - lines are not normalized so this style of move caused normalization changes.  I did only move the files.  There are no changes except pom.xml.  We should add .gitattributes so we normalize the code base.  It might even be advisable to do that before this PR is merged to be certain that I didn't change anything :)


MALfunction84:
Older versions of JBoss do require context.xml under both META-INF and WEB-INF. Check out the comments in #83 starting [here](https://github.com/psi-probe/psi-probe/issues/83#issuecomment-81729675) for the history. When we drop support in 3.0, this will be safe to remove. For now, removing it is likely to cause a regression.

Changing all line endings is akin to reformatting code; it completely breaks continuity for forkers and will cause merge conflicts. I agree that it needs to be done, but let's wait until right before we cut 2.4 loose and do it all at once.

If you can fix these two things, I can merge with a clear conscience.

hazendaz:
Do you know of a good way to change line endings?  It only showed as moves on my machine.  The changes showed after it his git.  I'll play around with normalization locally just to see how large of a problem this currently is.  I'm sure it's not isolated to just these few files.  With any luck I can get this fixed and back up including context.xml file tonight.

--- Original Message ---

MALfunction84:
I've always used Notepad++ for changing line endings, encoding, etc.

hazendaz:
I've tried various ways to get this working and nothing seems to work.  So let's just wait on this until either right after line normalization occurs or in 2.4.x.  I did check line normalization and it is pretty much everything.

I'm closing this for now.  I'll keep my branch and bring this back up later.

PR #522, opened 2015-06-28
Align tomcat poms 5, 6, 7, 8

hazendaz:
This has been sitting a bit on my machine...

Align poms for tomcat so they are fairly close to being the same.  Update to latest tomcat 7 and just one behind on tomcat 8.  Effectively not much going on here.

On jsp pages, language tag is obsolete and otherwise ignored.  Cleared all the warnings.  Again effectively not much going on here.

### DIFF
--- a/tomcat60adaptor/pom.xml
+++ b/tomcat60adaptor/pom.xml
@@ -36,29 +36,18 @@
 				</exclusions>
 			</dependency>
 			<dependency>
-				<groupId>javax.servlet</groupId>
-				<artifactId>servlet-api</artifactId>
-				<version>2.5</version>
-			</dependency>
-			<dependency>
 				<groupId>org.apache.tomcat</groupId>
 				<artifactId>catalina</artifactId>
-				<version>6.0.43</version>
-				<exclusions>
-					<exclusion>
-						<groupId>org.apache.tomcat</groupId>
-						<artifactId>servlet-api</artifactId>
-					</exclusion>
-				</exclusions>
+				<version>6.0.44</version>
 			</dependency>
 			<dependency>
 				<groupId>org.apache.tomcat</groupId>
 				<artifactId>jasper</artifactId>
-				<version>6.0.43</version>
+				<version>6.0.44</version>
 				<exclusions>
 					<exclusion>
-						<groupId>org.apache.tomcat</groupId>
-						<artifactId>servlet-api</artifactId>
+						<artifactId>ecj</artifactId>
+						<groupId>org.eclipse.jdt.core.compiler</groupId>
 					</exclusion>
 				</exclusions>
 			</dependency>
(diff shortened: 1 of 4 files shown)


MALfunction84:
I don't know what you mean by "align POMs." I agree with the stated goal--consistency is good--but I'm not sure I understand what specifically you're going for here. Can you provide some more detail?

Historically, I've been using ```<dependencyManagement>``` only in the parent pom. This way, the presence of ```<version>``` indicates that the dependency is specific to this artifact and not shared. Are there some practical benefits to abandoning this?

e7618cd should be a separate PR. [One PR per issue](https://github.com/psi-probe/psi-probe#contributions), please. Thanks!

hazendaz:
Goal here is making the poms for tomcat 5/6/7/8 match as closely as possible.  Only tomcat 5/6 were largely different so this just brings them in line.

As for dependency management, no previous pratice is being abandoned in this case.  This is how 7/8 were already setup.  There is good reasoning to doing this.  If you take a closer look you will notice that the size of the entire pom(s) dropped in line count quite a bit.  I think it otherwise makes it slightly more managable for us.  It keeps things consistent as best possible from 5 to 6 to 7 to 8.  That's normally my goal, less LOC.  However, it's real benefit is in how it handles transient dependencies.  And if there is a need to extend ever in the future, it's control is more fine tuned which should naturally mean less work and more predictable outcome.  It's far from perfect even in this sense for a number of reasons.  But again this was really just to mirror up close to tomcat 7/8.

I'll split out the other commit here shortly.  I've been so busy with other things I wanted to get this pushed off my plate but understand wanting separation.

hazendaz:
Tomcat 5/6 changes make it more in line with tomcat 7/8 setup.  Dependency Management can be used in same pom.  It controls the transient dependencies as well and isn't exactly perfect in maven terms.  I've seen it fail to work correclty a number of times.  However, the goal here was to try to make all 4 supported tomcats mirror each other as closely as possible.

hazendaz:
ok - top commit has been pulled off.  this is just the pom change now.  I'll push the other up later.  Thanks.

tomcat60adaptor/pom.xml
```diff
@@ -36,31 +36,20 @@
...
 				<artifactId>jasper</artifactId>
-				<version>6.0.43</version>
-				<exclusions>
-					<exclusion>
-						<groupId>org.apache.tomcat</groupId>
-						<artifactId>servlet-api</artifactId>
-					</exclusion>
-				</exclusions>
+				<version>6.0.44</version>
+			<exclusions>
```
MALfunction84: Can you fix the indentation here?

hazendaz:
rebased.

MALfunction84:
Indentation is still off in tomcat60adaptor/pom.xml.

hazendaz:
Indentation is fixed.  Bumpted tomcat 8 up to 23.  Still same number of commits.  Rebase was against master.  I've cleared all rebasing now including spring3 so the formatting is good.  May need to think about applying formatting to xml.  As far as I know that is a manual effort still.

PR #527, opened 2015-07-06
Fixes #505 Move Resources into proper location

### LINKED ISSUE #505: Resource bundles in non-standard location

hazendaz:


hazendaz:
Much cleaner now as this shows only moves and the two changes in the web/pom.xml.

Issue #17, opened 2015-03-16
Oracle JDBC driver is bound by distribution license


GoogleCodeExporter:
```
The Oracle JDBC driver, ojdbc14.jar, is bound by Oracle's distribution license:
http://www.oracle.com/technology/software/popup-license/distribution-license.htm
l

We need to answer these questions:
- How exactly does this license affect our work?
- Will Probe still work if we remove the .jar from the packaged .war?
- Should we remove the .jar from the repository, too?
  - If we need to remove it, we'll have to reset the repository and
re-import it without that .jar file to remove it from the history.
  - If we leave it, will the Oracle license have to apply to our source?
```

Original issue reported on code.google.com by `MALfunct...@gmail.com` on 23 Apr 2009 at 7:59

PR #528, opened 2015-07-06
Oracle local fixes #468

### LINKED ISSUE #468: Oracle Dependencies
Oracle dependencies appear to be targetting 11g for UCP and 10 for OJDBC.  Neither are available in central due to licensing restrictions.  I'm just jumping in here and builds fail as a result.  In order to make the build work as described on the landing README, it should store the jars directly in this project under a thirdParty directory with a bat file to load it into local maven.  I do this on a project for other needs already so I can port that over here.  That is unless I'm missing something here.

hazendaz:
While there is no perfect solution here, this internalizes the oracle dependencies so we do not need to rely on other repos.  This also allows users to stand behind a nexus repo that does not itself contain the oracle dependencies.


MALfunction84:
I could've sworn I've addressed this before. The Oracle dependencies are encumbered by the OTN license, which places limits on distribution. I can't assume the risk of hosting them here and violating the license. Already went through that once in #17.

hazendaz:
I do not feel this project violates the license.  The primary issues are export regulations, charging for the production, or touching their jars.  Given none of that applies, it is allowable for the build.  Further, there is only the need to make users agree to the licensing terms.  If we are straight by the book, what is being done now is a violation by proxy.

http://www.oracle.com/technetwork/licenses/distribution-license-152002.html

I think we solve all this by adding a section to the main page which should be their anyways as we are attributing to the jars getting to peoples machines in the build process regardless without any acknowledgement being made.  This is an idea.  Maybe this can be tweeked enough until this is acceptible.  Maybe it's even something we can pass by oracle legal department to ensure we are compliant.  Basically a huge disclaimer on the main README.

## License

To download, fork, or clone PSI-PROBE regardless of use-case, and regardless of which distribution, you must agree and accept the [OTN Standard License Terms](http://www.oracle.com/technetwork/licenses/standard-license-152015.html).  PSI-PROBE uses the oracle drivers only for building and further expects explicite agreement to the forthset license for any intended usage of the oracle drivers either by grabbing from the build or otherwise.  PSI-PROBE distribution makes no direct distribution of the oracle drivers and takes no liability for users violating terms oracle as set forth.

hazendaz:
One other thing.  I had the jmx pieces in this as well.  If those do not have strange licensing, then I can pull those out into a separate PR just on JMX pieces and the oracle discussion can continue otherwise.  I do think out of this the jmx piece and notification on the oracle license are both pluses even if the jars ultimately don't come into the process build.

MALfunction84:
Why hasn't Oracle put their drivers in Maven central? My guess is it's because of the license. IANAL, but I don't think we can mix GPLv2 with the OTN License. With respect, I don't find your feelings on the matter very convincing. Sorry, but I'm not willing to risk it. (Technically it's lds.org that's violating the license by distributing the oracle jars.)

As for JMX, I'm really not seeing any problems with the way it's set up. I realize it's a problem for your specific setup, as discussed in #468, but no one else has reported this problem. If the external repos aren't working for you, you have a workaround to install the artifacts locally.

PR #532, opened 2015-07-07
Fix listing global/application datasources on Tomcat 6 and 7

### LINKED ISSUE #531: Global/application datasources not listed on Tomcat 6 and 7
Dividing #490 to conquer. I fixed the Tomcat 6 and Tomcat 7 issue. Tomcat 8 is more complicated because of https://github.com/psi-probe/psi-probe/issues/490#issuecomment-105272331:

Tomcat 8 changes tomcat-dbcp package name from org.apache.tomcat.dbcp.dbcp to org.apache.tomcat.dbcp.dbcp2. So datasources don't show up because org.apache.commons.dbcp.BasicDataSource can't find the new package name.

@MALfunction84 Should I create a separated issue to Tomcat 8 or we use #490?

diogosantana:
Fixes #531.

The `AbstractTomcatContainer.bindToContext` and `unbindToContext` implementation doesn't work because it pass null to token.

I override both methods on Tomcat60ContainerAdaptor and Tomcat70ContainerAdaptor to pass the apropriate expected value.

I have test this on Tomcat 6 (Java 5 and 6) and on Tomcat 7 (Java 6), but I believe it will work on Java 7 and 8. I will test it on the next days.

### DIFF
--- a/tomcat60adaptor/src/main/java/com/googlecode/psiprobe/Tomcat60ContainerAdaptor.java
+++ b/tomcat60adaptor/src/main/java/com/googlecode/psiprobe/Tomcat60ContainerAdaptor.java
@@ -28,6 +28,7 @@ import org.apache.catalina.deploy.FilterDef;
 import org.apache.catalina.deploy.FilterMap;
 import org.apache.catalina.deploy.NamingResources;
 import org.apache.commons.modeler.Registry;
+import org.apache.naming.ContextBindings;
 import org.apache.naming.resources.Resource;
 import org.apache.naming.resources.ResourceAttributes;
 
@@ -401,4 +402,14 @@ public class Tomcat60ContainerAdaptor extends AbstractTomcatContainer {
     return result;
   }
 
+  @Override
+  public void bindToContext(Context context) throws NamingException {
+    ContextBindings.bindClassLoader(context, context,
+        Thread.currentThread().getContextClassLoader());
+  }
+
+  public void unbindFromContext(Context context) throws NamingException {
+    ContextBindings.unbindClassLoader(context, context,
+        Thread.currentThread().getContextClassLoader());
+  }
 }
--- a/tomcat70adaptor/src/main/java/com/googlecode/psiprobe/Tomcat70ContainerAdaptor.java
+++ b/tomcat70adaptor/src/main/java/com/googlecode/psiprobe/Tomcat70ContainerAdaptor.java
@@ -33,6 +33,7 @@ import org.apache.jasper.JspCompilationContext;
 import org.apache.jasper.Options;
 import org.apache.jasper.compiler.JspRuntimeContext;
 import org.apache.jasper.servlet.JspServletWrapper;
+import org.apache.naming.ContextBindings;
 import org.apache.naming.resources.Resource;
 import org.apache.naming.resources.ResourceAttributes;
 
@@ -436,4 +437,14 @@ public class Tomcat70ContainerAdaptor extends AbstractTomcatContainer {
     return result;
   }
 
+  @Override
+  public void bindToContext(Context context) throws NamingException {
+    ContextBindings.bindClassLoader(context, context,
+        Thread.currentThread().getContextClassLoader());
+  }
+
+  public void unbindFromContext(Context context) throws NamingException {
+    ContextBindings.unbindClassLoader(context, context,
+        Thread.currentThread().getContextClassLoader());
+  }
 }


robertoschwald:
Tested locally in TC 7.0.62. It works. Thanks!

robertoschwald:
Deployed a 2.4.0-SNAPSHOT with the fix to 3 servers. One strange thing I notice is, that in "Data Sources" I get "There are no data sources configured for this Tomcat instance", but when I click on an application with a jndi ds configured in context.xml, I can see the jndi datasource and click into it.

This happens on Tomcat-7.0.55 and Tomcat-7.0.42 servers here.

See Screenshots.

snap1

snap2


snap3

diogosantana:
Can you give more details. I will try to reproduce. Tell me Java version, jdbc driver, jdbc driver version and configuration of datasources (global and application).

robertoschwald:
- CentOS 6.6
 - Oracle Java SDK 1.7.0_79
 - Tomcat 7.0.55
 - mysql-connector-java-5.1.29.jar in tc lib dir

diogosantana:
I couldn't reproduce the error, but I'm on Windows 7 with Tomcat 7.0.62.
Can you debug on your machine?

@MALfunction84 Can you test it too?

diogosantana:
My screenshot:

image

robertoschwald:
Maybe it makes a difference if you got the context.xml file within the war in META-INF dir, or if you have it in <tomcat_home>/conf/Catalina/localhost/<app>.xml like we do. With the latter, it fails here.

Got the same effect here on 4 different servers running TC 7.0.55 and JDK 1.7.0_79.

diogosantana:
I am using the later. Can you debug?
Em 07/07/2015 09:00, "Robert Oschwald" <notifications@github.com> escreveu:

robertoschwald:
I debug today

robertoschwald:
While remote debugging in CentOS6.5 / JDK 7.0.79 / TC 7.0.55, I notice that  in TomcatJdbcPoolDatasouceAccessor, canMap() returns false for org.apache.tomcat.jdbc.pool.DataSource, because "resource instanceof DataSource" is false.

Very strange. Resource is of type: org.apache.tomcat.jdbc.pool.DataSource

If I comment out "&& resource instanceof DataSource" in canMap(), I get:

java.lang.ClassCastException: org.apache.tomcat.jdbc.pool.DataSource cannot be cast to org.apache.tomcat.jdbc.pool.DataSource
	at com.googlecode.psiprobe.beans.TomcatJdbcPoolDatasourceAccessor.getInfo(TomcatJdbcPoolDatasourceAccessor.java:36)

I have no idea why this happens...

robertoschwald:
Maybe a classloader issue. The resource class loader seems not to be the same as the TomcatJdbcPoolDatasourceAccessor ClassLoader:

```
DataSource Class Loader: 
WebappClassLoader@26103
  context: /cockpit
  delegate: false
  repositories:
    /WEB-INF/classes/
----------> Parent Classloader:
org.apache.catalina.loader.StandardClassLoader@63746939
```

```
TomcatJdbcDataSourceAccessor Class Loader:
WebappClassLoader@26143
  context: /probe
  delegate: false
  repositories:
    /WEB-INF/classes/
----------> Parent Classloader:
org.apache.catalina.loader.StandardClassLoader@63746939
```

robertoschwald:
Ok. It is a classLoader issue. As a workaround, I used reflection to invoke the resources's methods. It works then. I'm not sure if there is a better solution. I post the class here so you can think of it.

```
/*
 * Licensed under the GPL License. You may not use this file except in compliance with the License.
 * You may obtain a copy of the License at
 * 
 * http://www.gnu.org/licenses/old-licenses/gpl-2.0.html
 * 
 * THIS PACKAGE IS PROVIDED "AS IS" AND WITHOUT ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING,
 * WITHOUT LIMITATION, THE IMPLIED WARRANTIES OF MERCHANTIBILITY AND FITNESS FOR A PARTICULAR
 * PURPOSE.
 */
package com.googlecode.psiprobe.beans;

import com.googlecode.psiprobe.model.DataSourceInfo;

import org.apache.juli.logging.Log;
import org.apache.juli.logging.LogFactory;
import org.apache.tomcat.jdbc.pool.DataSource;

import java.lang.reflect.Method;

/**
 * Datasource accessor for the new <a
 * href="http://people.apache.org/~fhanik/jdbc-pool/jdbc-pool.html">Tomcat JDBC pool</a>.
 * 
 * <p>
 * Tomcat 6 can be configured to use the new pool for its datasources instead of the old DBCP-based
 * pool. SpringSource tc Server uses this new pool <a
 * href="http://vigilbose.blogspot.com/2009/03/apache-commons-dbcp-and-tomcat-jdbc.html">by
 * default</a>.
 * </p>
 *
 * @author chenwang
 * @author Mark Lewis
 */
public class TomcatJdbcPoolDatasourceAccessor implements DatasourceAccessor {
  private static Log log = LogFactory.getLog(TomcatJdbcPoolDatasourceAccessor.class);

  public DataSourceInfo getInfo(Object resource) throws Exception {

    DataSourceInfo dataSourceInfo = null;
    if (canMap(resource)) {
      dataSourceInfo = new DataSourceInfo();
      if (resource.getClass().getClassLoader() != this.getClass().getClassLoader()) {
        log.debug("Cast to DataSource of " + resource + " not possible. ClassLoaders different");
        Class clazz = resource.getClass();
        dataSourceInfo.setBusyConnections((Integer) clazz.getMethod("getNumActive").invoke(resource));
        dataSourceInfo.setEstablishedConnections((Integer)clazz.getMethod("getNumIdle").invoke(resource) + (Integer)clazz.getMethod("getNumActive").invoke(resource));
        dataSourceInfo.setMaxConnections((Integer)clazz.getMethod("getMaxActive").invoke(resource));
        dataSourceInfo.setJdbcURL((String) clazz.getMethod("getUrl").invoke(resource));
        dataSourceInfo.setUsername((String) clazz.getMethod("getUsername").invoke(resource));
        dataSourceInfo.setResettable(false);
        dataSourceInfo.setType("tomcat-jdbc");
      } else {
        DataSource source = (DataSource) resource;
        dataSourceInfo.setBusyConnections(source.getNumActive());
        dataSourceInfo.setEstablishedConnections(source.getNumIdle() + source.getNumActive());
        dataSourceInfo.setMaxConnections(source.getMaxActive());
        dataSourceInfo.setJdbcURL(source.getUrl());
        dataSourceInfo.setUsername(source.getUsername());
        dataSourceInfo.setResettable(false);
        dataSourceInfo.setType("tomcat-jdbc");
      }
    }
    return dataSourceInfo;
  }

  public boolean reset(Object resource) throws Exception {
    return false;
  }

  public boolean canMap(Object resource) {
    return resource.getClass().getName().equals("org.apache.tomcat.jdbc.pool.DataSource");
  }

}

```

jander99:
Shouldn't it be casting to a javax.sql.DataSource? Or am I reading what
you're attempting to do incorrectly?

robertoschwald:
Casting to javax.sql.DataSource doesn't help here, as the statistics methods (getNumActive, etc) are defined in o.a.t.jdbc.pool.DataSourceProxy, not in DataSource.

diogosantana:
My first thought is this should not happen. It seems to be more like a specific setup problem then a psi-probe problem.

Why you have another copy of `org.apache.tomcat.jdbc.pool.DataSource` inside your application? You shouldn't if I'm guessing right?

Once I answered a stackoverflow question with slf4j that might be helpful: http://stackoverflow.com/questions/29504180/slf4j-error-class-loader-have-different-class-objects-for-the-type/29504701#29504701

robertoschwald:
> Why you have another copy of org.apache.tomcat.jdbc.pool.DataSource inside your application? You shouldn't if I'm guessing right?

I'm not having a copy of it anywhere in my deployed applications or in probe.war. It is in the normal tomcat/lib/ jar. 

Configuration is not special at all here.
Plain tomcat installation from "zip" file with configured bin/setenv.sh and probe.war deployed to webapps directory. 
My applications do not reside in webapps directory but another dir at same directory level as "webapps". 

In tomcat/conf/Catalina/localhost/<application>.xml, the docBase points to 

docBase="${catalina.home}/applicationdeploy/application_name" directory. 

Maybe thats the culprit. It always was like this since Tomcat 5.x. Probe stopped working when we switched from TC 7.0.42 to TC 7.0.55 and the reason seems to be a change in ClassLoader isolation.

When using my reflection approach when classloaders are different, the ds get listed correctly.

Deployed my special variant (https://github.com/robertoschwald/psi-probe/tree/pr_532_via_reflection) to 6 servers now. All of them correctly display the datasources again.

diogosantana:
Ok. I understand that your variant is working, but I am not convinced that is the best solution. It is not clear to me what is the exact problem, because I tried to reproduce this behavior without success.

Perhaps it would be interesting set up a test case to reproduce the error in a more standard enviroment.

Can you check classloaders of the two org.apache.tomcat.jdbc.pool.DataSource just like you do in https://github.com/psi-probe/psi-probe/pull/532#issuecomment-119544655 ?

Maybe it will show the real problem...

MALfunction84:
@robertoschwald Reflection is informative but I agree with @diogosantana that it's not a good long-term solution. Is either of the classloaders you mentioned [earlier](#issuecomment-119544655) the same as PSI Probe's classloader? Both are WebappClassLoaders, so one of them is likely your app and the other is possibly PSI Probe's?

MALfunction84:
Found an answer to a [related question](http://stackoverflow.com/questions/19431102/loading-jdbc-driver-via-resource-tomcat-7) from markt himself. It's kind of late, so I'm going to leave it here and re-read it when I'm more awake.

robertoschwald:
That was a good hint, Mark. I checked my applications lib dir and saw that tomcat-jdbc is in there. Maybe thats the point. All of those applications we see the problem in psi-probe are Grails 2.4.4 applications, so I assume all of them got the tc libs deployed. I remove them from the build process and test again.

robertoschwald:
The Grails grails-plugin-datasource requires tomcat-jdbc, tomcat-embed-logging-log4j and tomcat-juli. Therefore, I can't remove them from the application WEB-INF/lib dir ..

So Tomcat seems to resolve from the application ClassLoader first and finds it in the Grails application. Is this assumption correct?

diogosantana:
Yes. You can change that behavior as described [here](https://tomcat.apache.org/tomcat-7.0-doc/class-loader-howto.html) (the `<Loader delegate="true"/>` part) and [here](https://tomcat.apache.org/tomcat-7.0-doc/config/loader.html).

robertoschwald:
checked by setting <Loader delegate="true"/> in one of the grails applications. With this, it works in psi-probe (with this PR merged), but that has side-effects on the application. Grails 2.4.4 datasource plugin depends on tomcat-jdbc-7.0.50 and is tested with this version. Tomcat 7.0.55 version is not tested with this library. Therefore, I strongly recommend to use the reflection approach as a fallback if classloaders differ. This way, nobody needs to tweak his applications to get Psi-Probe working.

diogosantana:
Does it work on tomcat 7.0.50?

robertoschwald:
Grails 2.4.4 recommended Tomcat Version is 7.0.55. https://grails.org/wiki/2.4.4%20Release%20Notes

diogosantana:
This is looking like a Grails issue rather then a Psi-Probe one. They are doing it wrong. grails-plugin-datasource should not rely on tomcat internal jars like tomcat-jdbc, or at least it should scope those provided.

Have you consider submit a issue to Grails?

diogosantana:
Did the `<Loader delegate="true"/>` set up work for you?

robertoschwald:
If adding to the Grails application, yes.

diogosantana:
To me is clear now this is not a Psi-Probe issue. It's a Grails one and it has a alternative solution based on Tomcat's feature made precisely to this type of situation.

robertoschwald:
Ack. I would add a note to the documentation that it won't work if tomcat-jdbc is in the applications classpath.

robertoschwald:
+1 for merging this PR.

Reference to the Grails ml post I sent there: 
https://groups.google.com/forum/#!topic/grails-dev-discuss/snwiCeDtXTA

jander99:
Can/should this PR include a section in the readme on how to enable the
grails workaround?

MALfunction84:
@jander99 The wiki is editable by all.

robertoschwald:
I've added wiki documentation to the troubleshooting section.

PR #541, opened 2015-07-31
Spring3 Upgrade

hazendaz:
Spring road to 4 needs a short pit stop in Spring 3 land.  spring-probe-resources.xml will need further review.  We have been running for a while now in production with this without issues.  Spring 4 is somewhat more complicated.  Once this has been thoroughly reviewed and any issues resolved, the work to spring 4 can start.

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -249,8 +249,23 @@
 			<scope>compile</scope>
 		</dependency>
 		<dependency>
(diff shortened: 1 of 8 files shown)


MALfunction84:
Looks good, and I agree with the approach. Since we're moving to Java 1.5, we can use the for-each style loops. I think I can get a bulk refactor going.

PR #557, opened 2015-10-26
List Tomcat DBCP datasources for versions before 8.x

ranophoenix:


MALfunction84:
Please make the following changes:
* Move the tomcat-dbcp dependency to `tomcat70adaptor/pom.xml`, and use the `tomcat.version` property defined there for its version.
* Move the `TomcatDbcpDatasourceAccessor` class to the tomcat70adaptor module, and rename it to `Tomcat7DbcpDatasourceAccessor`. (Note the addition of the 7.)
* Use the updated class name in `spring-probe-resources.xml`.
* [File an issue to track this change](https://github.com/psi-probe/psi-probe#contributions), and reference it in your PR.

Thanks!

PR #563, opened 2015-12-08
Added support for tomcat9 by copying and adjusting tomcat80adapter module

xomit:
Hi,

as I'm currently working and testing my applications with a Tomcat 9 instance, i've copied and
adjusted the tomcat80adapter, build and deployed the WAR to my tomcat instance and it seems
to work quite nice.

Though I have not tested the build intensively, maybe you will merge in the changes to your
master tree to start support for Tomcat 9.


hazendaz:
+1 to add this support.

Couple of things I would do and I'm sure the authors will want you to do.  Please comment out the main build pieces on this PR so that it does not impact our current setup and development.  Please note travis CI failed expectadly.  I did a similar thing on the waffle authentication framework so the code is there but it requires individual developers to build themselves.  I don't anticipate tomcat 9 will be out before we actually have the next release of this library anyways but I certainly feel very strongly this needs to come in as tomcat 9 is on the fast track now and we want to figure out each upgrade works as expected until it finally goes live.  So having it here but more or less silently supported I feel is entirely ok.  I would however lead to you for ensuring to keep pase of tomcat milestone and beta releases for any and all changes until such time they call that official.

MALfunction84:
Travis failed on OpenJDK 6, which is actually fine. We can build on a higher JDK and target the lower one. The troubling failure was on JDK 8. Looks like it just timed out.

I'm going to close and reopen to trigger a new build on Travis.

MALfunction84:
Removed openjdk6. Going to close and reopen one more time to rebuild.

Issue #379, opened 2015-03-16
Update quartz from 1.6.0 to 2.x


GoogleCodeExporter:
```
What steps will reproduce the problem?
1.login to Probe
2.enter login/password admin
3.

What is the expected result?
Access to Probe

What happens instead?
Application encountered an unexpected error :javax.servlet.ServletException: 
"Servlet.init()"
ERROR com.googlecode.psiprobe.ProbeServlet  - Context initialization failed
org.springframework.beans.factory.BeanCreationException: Error creating bean 
with name 'handlerMapping' defined in ServletContext resource 
[/WEB-INF/spring-probe-servlet.xml]: Initialization of bean failed; nested 
exception is org.springframework.beans.factory.CannotLoadBeanClassException: 
Error loading class [org.springframework.scheduling.quartz.CronTriggerBean] for 
bean with name 'connectorStatsTrigger' defined in ServletContext resource 
[/WEB-INF/spring-probe-stats.xml]: problem with class file or dependent class; 
nested exception is java.lang.IncompatibleClassChangeError: class 
org.springframework.scheduling.quartz.CronTriggerBean has interface 
org.quartz.CronTrigger as super class


What version of Probe are you using?
2.3.3

What environment (browser version, Tomcat version, JVM version, server OS)?
Windows 7 64bits, TomEE 1.5.1 (Tomcat 7.0.34), JDK 1.7.0_11-b21, FireFox 23/IE 
8.

Please provide any additional information below.
(Attach logs or stack traces as files instead of pasting the contents
here.)

```

Original issue reported on code.google.com by `dad.depo...@gmail.com` on 30 Oct 2013 at 4:59
* Blocked on: #449

Attachments:
* [probe.log](https://storage.googleapis.com/google-code-attachments/psi-probe/issue-379/comment-0/probe.log)

hazendaz:
In reference to users with this type of setup affected by older quartz.

For users using tomcat with mule ESB (anypoint) at the container level as well as using psi-probe for management will no longer by able to use psi probe with mule 3.7.0.  Mulesoft upgraded quartz which breaks on a casting error.  Only solution is to get psi probe upgraded to use quartz 2.2.1 in this case.

MALfunction84:
Yep, and we can't update quartz until we update Spring to 3.x.

hazendaz:
[...]

PR #567, opened 2015-12-20
[#379 Fix] Update to latest quartz 2.2.2 with spring and slf4j adjust…

hazendaz:
…ments.  Quartz 2 uses slf4j so reworked it a bit so logging from that library will go to log4j as intended in the current psi-probe setup.


hazendaz:
Fixes #379

PR #588, opened 2016-01-18
Per Sonar analysis, add missing libraries and correct java mail usage - Fixes #587

### LINKED ISSUE #587: Add libraries reported as missing for sonar analysis
Sonar analysis shows that the following libraries are missing from analysis.

tomcat-coyote scope provided
jsr305 scope provided
java mail requires implementation scope compile

hazendaz:
Java mail doesn't come with tomcat so usage of it requires we utilize
the implementation and api not just the api.

This fixes #587

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -32,6 +32,12 @@
 		</dependency>
 		<dependency>
 			<groupId>org.apache.tomcat</groupId>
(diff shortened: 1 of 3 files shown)


hazendaz:
coveralls seems to not be responding today.  So merging since travis-ci was all good.

PR #595, opened 2016-01-18
Java Language Cleanups - Remove unnecessary else statements

hazendaz:
Per eclipse, performing various language cleanups.  As a direct result, some issues do get removed from sonar but very few.

### DIFF
--- a/core/src/main/java/com/googlecode/psiprobe/beans/ContainerWrapperBean.java
+++ b/core/src/main/java/com/googlecode/psiprobe/beans/ContainerWrapperBean.java
@@ -99,9 +99,8 @@ public class ContainerWrapperBean {
                   tomcatContainer = (TomcatContainer) obj;
                   tomcatContainer.setWrapper(wrapper);
                   break;
-                } else {
-                  logger.debug("Cannot bind " + className + " to " + serverInfo);
                 }
+                logger.debug("Cannot bind " + className + " to " + serverInfo);
               } else {
                 logger.error(className + " does not implement " + TomcatContainer.class.getName());
               }
--- a/core/src/main/java/com/googlecode/psiprobe/tokenizer/Tokenizer.java
+++ b/core/src/main/java/com/googlecode/psiprobe/tokenizer/Tokenizer.java
@@ -276,9 +276,8 @@ public class Tokenizer {
           if (compare(symbol.startText.toCharArray(), 1)) {
             result = index;
             break;
-          } else {
-            index++;
           }
+          index++;
         } else {
           break;
         }
--- a/core/src/main/java/com/googlecode/psiprobe/beans/ResourceResolverBean.java
+++ b/core/src/main/java/com/googlecode/psiprobe/beans/ResourceResolverBean.java
@@ -234,9 +234,8 @@ public class ResourceResolverBean implements ResourceResolver {
 
       if (obj instanceof DataSource) {
         return (DataSource) obj;
-      } else {
-        return null;
       }
+      return null;
     } finally {
       if (context != null) {
         containerWrapper.getTomcatContainer().unbindFromContext(context);
--- a/core/src/main/java/com/googlecode/psiprobe/beans/stats/listeners/ThresholdListener.java
+++ b/core/src/main/java/com/googlecode/psiprobe/beans/stats/listeners/ThresholdListener.java
@@ -137,12 +137,11 @@ public abstract class ThresholdListener extends AbstractStatsCollectionListener 
           + " is not defined or inherited.  Disabling listener for \"" + name + "\" series.");
       setSeriesDisabled(name, true);
       return DEFAULT_THRESHOLD;
-    } else {
-      try {
-        return SizeExpression.parse(threshold);
-      } catch (NumberFormatException ex) {
-        return DEFAULT_THRESHOLD;
-      }
+    }
+    try {
+      return SizeExpression.parse(threshold);
+    } catch (NumberFormatException ex) {
+      return DEFAULT_THRESHOLD;
     }
   }
 
--- a/core/src/main/java/com/googlecode/psiprobe/controllers/ContextHandlerController.java
+++ b/core/src/main/java/com/googlecode/psiprobe/controllers/ContextHandlerController.java
@@ -39,16 +39,15 @@ public abstract class ContextHandlerController extends TomcatContainerController
 
     if (context != null || isContextOptional()) {
       return handleContext(contextName, context, request, response);
-    } else {
-      if (contextName != null) {
-        request.setAttribute(
-            "errorMessage",
-            getMessageSourceAccessor().getMessage("probe.src.contextDoesntExist",
-                new Object[] {contextName}));
-      }
-
-      return new ModelAndView("errors/paramerror");
     }
+    if (contextName != null) {
+      request.setAttribute(
+          "errorMessage",
+          getMessageSourceAccessor().getMessage("probe.src.contextDoesntExist",
+              new Object[] {contextName}));
+    }
+
+    return new ModelAndView("errors/paramerror");
   }
 
   /**
--- a/core/src/main/java/com/googlecode/psiprobe/controllers/ErrorHandlerController.java
+++ b/core/src/main/java/com/googlecode/psiprobe/controllers/ErrorHandlerController.java
@@ -95,8 +95,7 @@ public class ErrorHandlerController extends AbstractController {
     String originalUri = (String) request.getAttribute("javax.servlet.error.request_uri");
     if (originalUri != null && originalUri.endsWith(ajaxExtension)) {
       return new ModelAndView(ajaxViewName);
-    } else {
-      return new ModelAndView(viewName);
     }
+    return new ModelAndView(viewName);
   }
 }
--- a/core/src/main/java/com/googlecode/psiprobe/controllers/jsp/DisplayJspController.java
+++ b/core/src/main/java/com/googlecode/psiprobe/controllers/jsp/DisplayJspController.java
@@ -51,9 +51,8 @@ public class DisplayJspController extends ContextHandlerController {
     if (compile) {
       return new ModelAndView(new RedirectView(request.getRequestURI() + "?webapp="
           + (contextName.length() == 0 ? "/" : contextName)));
-    } else {
-      return new ModelAndView(getViewName(), "summary", summary);
     }
+    return new ModelAndView(getViewName(), "summary", summary);
   }
 
 }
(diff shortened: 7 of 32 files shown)


jander99:
These all look good, except I would prefer that maybe commit 1ab7251 be spun off to its own PR.  While its a cleanup commit, you're also changing debug in some places.  Let's review that separately so we can reinforce some better logging where it's needed.

hazendaz:
@jander99 I'll spit this into two PRs.  I'm done for the night coding so will be tomorrow.

jander99:
@hazendaz Fine by me, the email spam has been a welcome change though, that's for sure!  You deserve a break and a cold beverage of your choosing.

hazendaz:
@jander99 that is my plan :)  @MALfunction84 is wanting to get in touch with you.  Do you have a personal email he can get in touch with you?

jander99:
jander99@gmail.com

hazendaz:
thanks

core/src/main/java/com/googlecode/psiprobe/beans/ContainerWrapperBean.java
```diff
@@ -99,9 +99,8 @@ public void setWrapper(Wrapper wrapper) {
                   tomcatContainer = (TomcatContainer) obj;
                   tomcatContainer.setWrapper(wrapper);
                   break;
-                } else {
-                  logger.debug("Cannot bind " + className + " to " + serverInfo);
                 }
+                logger.debug("Cannot bind " + className + " to " + serverInfo);
```
MALfunction84: This is more of a style thing. I wouldn't call this necessary, but I don't oppose it.

core/src/main/java/com/googlecode/psiprobe/beans/stats/providers/MultipleSeriesProvider.java
```diff
@@ -36,7 +36,7 @@
 public class MultipleSeriesProvider extends AbstractSeriesProvider {
   
   /** The stat name prefix. */
-  private String statNamePrefix;
```
MALfunction84: Why remove this modifier? I tend to make everything private unless there's a compelling reason not to.

core/src/main/java/com/googlecode/psiprobe/tokenizer/Tokenizer.java
```diff
@@ -267,18 +267,17 @@ private int lookupSymbol(char chr) throws IOException {
     if (index >= 0) {
       // the index could be anywhere within a group of sybols with the same first letter
       // so we need to scroll up the group to make sure we start test from the beginning
-      while (index > 0 && ((TokenizerSymbol) symbols.get(index - 1)).compareTo(chrObj) == 0) {
+      while (index > 0 && symbols.get(index - 1).compareTo(chrObj) == 0) {
         index--;
       }
       while (index < symbols.size()) {
-        TokenizerSymbol symbol = ((TokenizerSymbol) symbols.get(index));
+        TokenizerSymbol symbol = (symbols.get(index));
```
MALfunction84: Extra parentheses can be removed.

core/src/main/java/com/googlecode/psiprobe/tokenizer/Tokenizer.java
```diff
@@ -168,7 +168,7 @@ public Token nextToken() throws IOException {
           // we have found a symbol
           TokenizerToken workToken =
               token.type == Tokenizer.TT_TOKEN && token.text.length() > 0 ? upcomingToken : token;
-          TokenizerSymbol symbol = ((TokenizerSymbol) symbols.get(symbolIndex));
+          TokenizerSymbol symbol = (symbols.get(symbolIndex));
```
MALfunction84: Extra parentheses can be removed.

core/src/main/java/com/googlecode/psiprobe/tokenizer/Tokenizer.java
```diff
@@ -90,7 +90,7 @@ public Tokenizer(Reader reader) {
    * @param cacheBufferSize the cache buffer size
    */
   public Tokenizer(Reader reader, int cacheBufferSize) {
-    symbols = new UniqueList();
+    symbols = new UniqueList<TokenizerSymbol>();
```
MALfunction84: Does this still work? Tests?

core/src/main/java/com/googlecode/psiprobe/tools/Whois.java
```diff
@@ -142,16 +142,16 @@ public static Response lookup(String server, int port, String query, long timeou
   public static class Response {
 
     /** The summary. */
-    private String summary;
```
MALfunction84: Why not private?

core/src/main/java/com/googlecode/psiprobe/tools/Whois.java
```diff
@@ -142,16 +142,16 @@ public static Response lookup(String server, int port, String query, long timeou
...
+    String summary;
     
     /** The data. */
-    private Map<String, String> data = new TreeMap<String, String>();
```
MALfunction84: Why not private?

core/src/main/java/com/googlecode/psiprobe/tools/Whois.java
```diff
@@ -142,16 +142,16 @@ public static Response lookup(String server, int port, String query, long timeou
...
+    Map<String, String> data = new TreeMap<String, String>();
     
     /** The server. */
-    private String server;
```
MALfunction84: Why not private?

core/src/main/java/com/googlecode/psiprobe/tools/Whois.java
```diff
@@ -142,16 +142,16 @@ public static Response lookup(String server, int port, String query, long timeou
...
+    String server;
     
     /** The port. */
-    private int port;
```
MALfunction84: Why not private?

core/src/test/java/com/googlecode/psiprobe/jsp/VisualScoreTagTest.java
```diff
@@ -70,7 +76,7 @@ private void doTestRangeScan(int fullBlocks, int partialBlocks, boolean invertLo
       }
     }
     if (count > 0) {
-      fail("Incorrect values were founded " + count + " times");
+      Assert.fail("Incorrect values were founded " + count + " times");
```
MALfunction84: I often do a static import for fail().

core/src/main/java/com/googlecode/psiprobe/beans/stats/providers/MultipleSeriesProvider.java
hazendaz: Inner class accesses the field.  If Eclipse checks are turned on, it will flag this condition wanting it to be default scoped.  Generally I try to avoid inner classes, primarily becuase they are really hard to test successfully.  I think a better fix depending on the condition is extracting the inner class.  I also try to make everything private as I find it more confusing otherwise but this is the easiest fix for now.  I think it does more to point out code smell issues more than anything else.

core/src/test/java/com/googlecode/psiprobe/tools/WhoisTests.java
```diff
@@ -23,7 +22,7 @@
  *
  * @author Mark Lewis
  */
-public class WhoisTests extends TestCase {
+public class WhoisTests {
```
MALfunction84: Should probably `@Ignore` this test and uncomment lines 43-44.

MALfunction84:
Love all the generics I'm seeing here. I clearly missed some in a112103.

core/src/main/java/com/googlecode/psiprobe/tokenizer/Tokenizer.java
hazendaz: I'll catch those on another commit.  There are quite a few of those that Sonar picked up on.  I left alone here so I could isolate the change occurring better.

jander99: I was thinking of building a Jenkins/sonar vm on DigitalOcean for fun.  I'm
cool right?  I might use psi probe as a test project for it.

hazendaz:
Eclipse can do a really good job of flushing these out.  However, the default eclipse doesn't have much of anything turned on.  I presume that is due to legacy concerns as it can be quite overwhelming.  At the moment, even after these, Eclipse is still complaining about close to 2000 issues.  The bulk of them being qualification and property hiding issues.  Most likely a future resolution.

core/src/main/java/com/googlecode/psiprobe/tokenizer/Tokenizer.java
hazendaz: That sounds great!  Last I checked on sonar, there was a pay for use version out there.  It's probably been a year.  I think it was rather cost prohibitive at the time.  I did just quickly look at this.  Sounds like this might be a cheap alternative.  If sonar is accessible, it can be tied into Pull Requests as another check before merging.

core/src/main/java/com/googlecode/psiprobe/beans/ContainerWrapperBean.java
hazendaz: I believe there is a little more to it.  Basically, the conditional as a whole is not completely intact because it breaks out of the conditional.  So one Eclipse check and I think others state this is bad.  Of course those checkers go on further to state that we shouldn't have multiple exit points (ie break, continue, return, etc).  One step at a time.  I don't think Eclipse goes far enough though as it only mindlessly suggests these fixes.  I know Intellij sort of layers the changes to the point the developer actually learns something in the process.  At the moment, I'm just taking the mindless approach :)  Later, deep diving using Intellij will become necessary.

MALfunction84: Well, this is only triggering when the `if` block has `return` or `continue` or `break` or any other flow control statement. In those cases, whether what follows is in an `else` block or not really doesn't matter. The flow will be identical. That's all I mean. You could argue that either way is fragile.

Eclipse is very opinionated. I wouldn't get too caught up trying to fix everything the warn you about.

core/src/main/java/com/googlecode/psiprobe/beans/stats/providers/MultipleSeriesProvider.java
MALfunction84: Ok, makes sense. The inner class is private too. No harm in maintaining privacy, but like I said, Eclipse is very opinionated.

I'm fine with inner classes that encapsulate something very specific, but I can understand the desire to move it out to maybe a package-private class for testing purposes. Still, the tests on a containing class *should* exercise its inner classes completely.

hazendaz:
@jander99 @MALfunction84 

Remove all commits but the removal of unnecessary 'if' statement.  I'll bring in others in smaller batches as well so it's easier to follow the single change.  Please review and feel free to merge.

hazendaz:
my bad...didn't mean to merge.  All checks passed though and this was otherwise thoroughly reviewed last night.  Please let me know if anything looks out of order here.

PR #601, opened 2016-01-23
Java Language Cleanups - Micro Updates

hazendaz:
Prevent Synthetic Accessor by adding constructor
Add empty arrays as [0] instead of [] {} for example new URL[0] instead of new URL[] {}
Remove Unnecessary constructors
Comment required contructors where constructors are overridden

### DIFF
--- a/core/src/main/java/com/googlecode/psiprobe/tools/AccessorFactory.java
+++ b/core/src/main/java/com/googlecode/psiprobe/tools/AccessorFactory.java
@@ -19,11 +19,6 @@ package com.googlecode.psiprobe.tools;
 public class AccessorFactory {
 
   /**
-   * Instantiates a new accessor factory.
-   */
-  private AccessorFactory() {}
-
-  /**
    * Gets the single instance of AccessorFactory.
    *
    * @return single instance of AccessorFactory
--- a/core/src/main/java/com/googlecode/psiprobe/tools/SecurityUtils.java
+++ b/core/src/main/java/com/googlecode/psiprobe/tools/SecurityUtils.java
@@ -29,11 +29,6 @@ import javax.servlet.http.HttpServletRequest;
 public class SecurityUtils {
 
   /**
-   * Instantiates a new security utils.
-   */
-  private SecurityUtils() {}
-
-  /**
    * Checks for attribute value role.
    *
    * @param servletContext the servlet context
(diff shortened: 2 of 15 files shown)


core/src/main/java/com/googlecode/psiprobe/tools/AccessorFactory.java
```diff
@@ -19,11 +19,6 @@
 public class AccessorFactory {
 
   /**
-   * Instantiates a new accessor factory.
-   */
-  private AccessorFactory() {}
```
MALfunction84: Utility classes with only static methods often have private constructors to prevent them from being instantiated. Why remove them?

hazendaz: Good catch!  My mistake.  Sonar likely flagged me.  Class needs to be final as well.  I'll fix that.  Thanks.  Appears there are a couple of these.  Sorry about that.  Sonars first complaint was not documenting them.  I thought they were all public.  Fix will come in next couple of days once you get a chance to look over the big pull request sitting out there :)

hazendaz: Sonar might not actually be complaining on this since these are not 'static final'.  I'm going to review real quick to see where the breakdown is.  I still should have seen it when looking.

core/src/main/java/com/googlecode/psiprobe/tools/SecurityUtils.java
```diff
@@ -29,11 +29,6 @@
 public class SecurityUtils {
 
   /**
-   * Instantiates a new security utils.
-   */
-  private SecurityUtils() {}
```
MALfunction84: Same goes for this private constructor.

core/src/main/java/com/googlecode/psiprobe/tools/AccessorFactory.java
MALfunction84: Makes sense to add `final` to the class and methods.

core/src/main/java/com/googlecode/psiprobe/tools/SecurityUtils.java
hazendaz: PR is coming with this fix.  I did review my JavaBeanTester code to see what was otherwise causing failure.  It was not internals for testing no-arg constructor.  It was actually kicking off a bug.  So I filed an issue to self there and will ignore related test case.

core/src/main/java/com/googlecode/psiprobe/tools/AccessorFactory.java
hazendaz: PR is coming with this fix.  I did review my JavaBeanTester code to see what was otherwise causing failure.  It was not internals for testing no-arg constructor.  It was actually kicking off a bug.  So I filed an issue to self there and will ignore related test case.

Rather than introduce even more changes in this area, just keeping it simple as originally intended.

PR #607, opened 2016-01-24
Add javabean tester and tests

hazendaz:
Javabean-tester is a project I own which allows for use of builder pattern to quickly test javabeans (ie POJOs).  Most projects have lots of these including this project.  Based on the low code coverage that currently exists, this is a key way to greatly improve code coverage very quickly and easily.  This has been done across the code base.  Getters and setters are internally verfied that when value is set, the get of that property will return the expected value used in setting.  This guarentees the getters and setters are written correctly.  The only area I skipped adding this was where code coverage already existed.  However, later it might make sense based on how thorough this is to add in those locations as well.

### DIFF
--- a/core/src/main/java/com/googlecode/psiprobe/tools/url/UrlParser.java
+++ b/core/src/main/java/com/googlecode/psiprobe/tools/url/UrlParser.java
@@ -34,6 +34,13 @@ public class UrlParser {
 
   /**
    * Instantiates a new url parser.
+   */
+  public UrlParser() {
+    // Required due to override
+  }
+
+  /**
+   * Instantiates a new url parser.
    *
    * @param url the url
    * @throws MalformedURLException the malformed url exception
--- a/core/src/main/java/com/googlecode/psiprobe/tools/MailMessage.java
+++ b/core/src/main/java/com/googlecode/psiprobe/tools/MailMessage.java
@@ -48,6 +48,13 @@ public class MailMessage {
 
   /**
    * Instantiates a new mail message.
+   */
+  public MailMessage() {
+    // Require Due To Override
+  }
+
+  /**
+   * Instantiates a new mail message.
    *
    * @param to the to
    * @param subject the subject
--- a/core/src/test/java/com/googlecode/psiprobe/beans/ContainerWrapperBeanTest.java
+++ b/core/src/test/java/com/googlecode/psiprobe/beans/ContainerWrapperBeanTest.java
@@ -1,0 +1,20 @@
+package com.googlecode.psiprobe.beans;
+
+import org.junit.Test;
+
+import com.codebox.bean.JavaBeanTester;
+
+/**
+ * The Class ContainerWrapperBeanTest.
+ */
+public class ContainerWrapperBeanTest {
+
+  /**
+   * Javabean tester.
+   */
+  @Test
+  public void javabeanTester() {
+    JavaBeanTester.builder(ContainerWrapperBean.class).loadData().test();
+  }
+
+}
--- a/core/src/test/java/com/googlecode/psiprobe/beans/stats/collectors/AppStatsCollectorBeanTest.java
+++ b/core/src/test/java/com/googlecode/psiprobe/beans/stats/collectors/AppStatsCollectorBeanTest.java
@@ -1,0 +1,20 @@
+package com.googlecode.psiprobe.beans.stats.collectors;
+
+import org.junit.Test;
+
+import com.codebox.bean.JavaBeanTester;
+
+/**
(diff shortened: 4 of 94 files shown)


core/src/main/java/com/googlecode/psiprobe/tools/url/UrlParser.java
```diff
@@ -34,6 +34,13 @@
 
   /**
    * Instantiates a new url parser.
+   */
+  public UrlParser() {
```
MALfunction84: This was purposefully non-public. If you ever instantiate this class with this constructor, it's going to spit out a bunch of blank values.

hazendaz: Learning...ok sonar normally complains about this and I've been adding these for years.  Ran some quit tests, did a little more research, and now certainly see what this is written this way.  The default one is junk anyways even with my previous knowledge and would likely fail.  What I didn't know was that java wouldn't otherwise allow creation.  I assumed wrongly that it did and adding it was just for clarity sake.  So I'm going to run some more scans to sonar without to see where it complains and see if it doesn't provide a more noval solution.

MALfunction84: I hate to ask, but do you mind going back through all the recent no-arg constructor additions and removals?

hazendaz: Yes, I will go back through all those.  In fact, I’m going to circle back and review everything recently done in last few days.  I’ve learned a couple of things here I didn’t know so I want to make sure I apply those correctly. 

 

I also reran some sonar checks and issues it use to complain about are not there now.  So obviously they were wrong at some point in time.  I’ve gotten so used to doing certain things that I take for granted what is right sometimes.

 

I’m getting ready to jump off for the night.  If possible, can you open a new issue with bullet points on what you want me to specifically look into (in case there are more)?  That might be useful in tracking it in case I get pulled off into life the next couple of days.  At the very least I don’t want to miss getting these back into their current state by Wednesday or so.

 

On the bright side, it was probably a wise decision to pull these in smaller chunks :)

 

Thanks,

 

Jeremy

MALfunction84: I reluctantly suggest you revert the recent changes and take it slower. Maybe add tests for one class at a time so it's not so much to review at once. They should be relatively independent anyway, I'd think.

hazendaz: You are right.  I should have avoided changing source code when adding these tests in general at this level.  There were a number I couldn't get working anyways so I added ```@ignore``` on them but I didn't in this case based on poor knowledge.  I should have done that with these constructor changes as well.  It certainly would have been easier to bring this in with zero source changes so review wasn't necessary then make adjustments later for additional support.  I don't know that i want to hit the revert button but I certainly will go back through all these in next couple of days and revert the specific source code changes in question.  I'll do that prior to pulling in the packaging changes.

Thanks for reviewing all this stuff.  Other than questionable changes, so far so good.  I'll get a little more deeper testing over next couple of days as well.  Do you know if there are use cases out there presently where people are extenting this library other than drop and go?

hazendaz: I opened #617 to address concerns.

MALfunction84: Yeah, the vastly increased test coverage is a welcome sight! We'll get this ironed out. Thanks for being open to revisiting these changes.

I don't think anyone is extending it or depending on it. The only intended client for its core API is the web app, and then much of the API doesn't work outside the context of the PSI Probe web app running on a supported server. Out of curiosity, why do you ask?

hazendaz: -- Proper location for this message, sorry for noise --

PR is coming with this fix.  I did review my JavaBeanTester code to see what was otherwise causing failure.  It was not internals for testing no-arg constructor.  It was actually kicking off a bug.  So I filed an issue to self there and will ignore related test case.

hazendaz: Answer regarding if anyone was using our internals was more of how much impact something like this might cause.  Basically taking that into consideration.  After thorough review, I feel issues were minor but definitely worth catching in a code review so they are not introduced and later expected behavior even if coded in error.  Anyways, all good here.  Thanks.

PR #608, opened 2016-01-24
Cleanup unnecessary setting of properties that jvm does by default

hazendaz:


core/src/main/java/com/googlecode/psiprobe/model/TomcatTestReport.java
```diff
@@ -27,8 +27,8 @@
   /** The Constant TEST_FAILED. */
   public static final int TEST_FAILED = 2;
 
-  /** The default memory size. */
-  private final int defaultMemorySize = 1024 * 1024; // 1MB
+  /** The default memory size. 1 MB */
```
MALfunction84: Better to use `{@value #defaultMemorySize}`

hazendaz: I switched to that just now but it doesn't dispaly like I would like.  Instead, if I list @value 1MB, it looks more reasonable in the javadoc.  If I understand correctly, the #defaultMemorySize should show the value of the field but doesn't seem to.  Any other suggestions?

Here is what I have at this point

```java
  /**
   * The default memory size. 
   *
   * @value 1MB
   */
  private final int defaultMemorySize = 1024 * 1024;
```

MALfunction84: Oh, looks like the field needs to be `static` too. Of course, it will show up as a 7-digit number...

hazendaz: PR is coming with this addressed.  It was a little more complicated but done. The getter had to go, static is public, and now reference directly.

PR #612, opened 2016-01-24
Use logback instead of log4j 1 - Fixes #611

### LINKED ISSUE #611: Migrate from log4j 1 as native usage to Logback
Reason.

Log4j1 is obsolete.

Choices moving forwards.

Log4j 2 or Logback

Current state of third parties we use.

Mostly have dependency on Slf4j.  None are using log4j2.

Solution.

Use logback natively since it is the native implementation for Slf4j and fully satisfies are needs.

This will remove slf4j-to-log4j and replace commons-logging with jcl-over-slf4j

hazendaz:
This pull request will switch native logging from log4j 1 to logback.  Given we have dependencies on either commons logging or slf4j, it makes sense to use logback since it natively implements slf4j.  Commons logging has next to no hit on switching as it just fakes out usage.  This will enhance the usage of quartz that was using slf4j log4j12.  Plus logback is heavily imporved of log4j 1.  The change is rather simple.  Attached is output of startup using the newly configured setup.  Please review and let me know if this works out for everyone.

[probe.txt](https://github.com/psi-probe/psi-probe/files/102198/probe.txt)

Fixes issue #611


hazendaz:
I'm going to merge this in so I can bring in a bigger change effectively switching library off googlecode and allowing for maven central releases.  I'm going to open a separate ticket to switch over the commons-logging we were otherwise using to slf4j.  That will come in right after the major switch over in order to prevent mixing the two.  If any additional changes such as adjustments to the logging style are necessary, please make note on the new issue I open.  Thanks.

PR #623, opened 2016-01-26
Remove tomcat 7.0.15 and lower support - Fixes #616

### LINKED ISSUE #616: Drop support of tomcat 7.0.15 and lower
Per code snippet

```Java
      /*
       * The above constructor's signature changed in Tomcat 7.0.16:
       * http://svn.apache.org/viewvc?view=revision&revision=1124719
       * 
       * If we reach this point, we are running on a prior version of Tomcat 7 and must use
       * reflection to create this object.
       */
      try {
        jcctx =
            ConstructorUtils.invokeConstructor(JspCompilationContext.class,
                new Object[] {name, false, opt, sctx, null, jrctx}, new Class[] {String.class,
                    Boolean.TYPE, Options.class, ServletContext.class, JspServletWrapper.class,
                    JspRuntimeContext.class});
        return jcctx;
```
We should not support known vulnerable versions of tomcat.  Anyone using such versions doesn't care about security and therefore can use psi probe 2.4.0.  Solutions like these are extremely hacky.  If we need to do things like this we should release parallel supported version and not mix legacy support and new support as a general rule of thumb.  At this point though, 7.0.16 was released almost 4 years ago so supporting it makes little to no sense.

hazendaz:
Remove support for tomcat 7.0.15 as that poses security risk in general usage and assumes we are supporting that version of tomcat.  This fixes issue #616

### DIFF
--- a/tomcat70adapter/src/main/java/com/googlecode/psiprobe/Tomcat70ContainerAdapter.java
+++ b/tomcat70adapter/src/main/java/com/googlecode/psiprobe/Tomcat70ContainerAdapter.java
@@ -24,17 +24,14 @@ import org.apache.catalina.deploy.ContextResourceLink;
 import org.apache.catalina.deploy.FilterDef;
 import org.apache.catalina.deploy.FilterMap;
 import org.apache.catalina.deploy.NamingResources;
-import org.apache.commons.beanutils.ConstructorUtils;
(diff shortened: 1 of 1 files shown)


jander99:
This matches with the Tomcat 8.0 container, so should be all we need.

PR #630, opened 2016-01-27
Fix line endings on recently added test case

hazendaz:


hazendaz:
This was causing issue on everything I did so just merged it so the issue goes away.

PR #633, opened 2016-01-27
Fix sift appender - Fixes #432

### LINKED ISSUE #432: Logback SiftingAppender broken in Logback v1.1.x
```
1.Use Logback Version 1.X.X (such as 1.1.1)
2.Use a Sifting Appender in your logback.xml
3.Deploy Probe and Click on the 'Logs' Tab after the Application has started
4.In the probe.log you'll find an Exception

What is the expected result?


What happens instead?
java.lang.NoSuchMethodException: No such accessible method: valueList() on 
object: ch.qos.logback.core.sift.AppenderTracker


2.3.3


(Attach logs or stack traces as files instead of pasting the contents
here.)

```

Original issue reported on code.google.com by `patrick....@gmail.com` on 29 Nov 2014 at 12:46

Attachments:
* [Fixed_Logback_Sifting_Appender_Support_for_Logback_Version___1_x_x.patch](https://storage.googleapis.com/google-code-attachments/psi-probe/issue-432/comment-0/Fixed_Logback_Sifting_Appender_Support_for_Logback_Version___1_x_x.patch)

hazendaz:
Fixes #432 See issue for details of testing and research on issue.

### DIFF
--- a/core/src/main/java/psiprobe/tools/logging/logback/LogbackLoggerAccessor.java
+++ b/core/src/main/java/psiprobe/tools/logging/logback/LogbackLoggerAccessor.java
@@ -146,7 +146,12 @@ public class LogbackLoggerAccessor extends DefaultAccessor {
     if ("ch.qos.logback.classic.sift.SiftingAppender".equals(appender.getClass().getName())) {
       Object tracker = MethodUtils.invokeMethod(appender, "getAppenderTracker", null);
       if (tracker != null) {
-        return (List<Object>) MethodUtils.invokeMethod(tracker, "valueList", null);
+        try {
+          return (List<Object>) MethodUtils.invokeMethod(tracker, "allComponents", null);
+        } catch (final NoSuchMethodException e) {
+            // XXX Legacy 1.0.x and lower support for logback
+            return (List<Object>) MethodUtils.invokeMethod(tracker, "valueList", null);
+        }
       }
       return new ArrayList<Object>();
     }


MALfunction84:
Can you maintain backwards-compatibility? There is a patch on the issue that attempts to do so. It's one of the reasons these logging components use reflection.

hazendaz:
Personal stance is that the request to fix is already coming up on a year and the change it intends to fix occurred nearly 3 years ago with official live date 2 years ago.  

If the logback change recently occurred, I can see continued support.  I think valid 2 year support window makes sense.  With this boarding on being a faily old issue, I don't see a need plus I sort of doubt SIFT appender is used in majority of cases.  However, if added back, I'd ask that we do so marking it deprecated and provide a date or version we believe we will remove the code.  If that is the direction, I can work on getting that added in here tomorrow night.

MALfunction84:
You and I have very different support models. :) It seems a very low complexity cost to continue supporting what's still called logback 1.x. (It's unfortunate that logback didn't reflect changes to their public API by bumping up the major version number.)

hazendaz:
I'll get this back in tonight.


Sent from Outlook Mobile






On Tue, Jan 26, 2016 at 9:11 PM -0800, "Mark Lewis" <notifications@github.com> wrote:





You and I have very different support models. :) It seems a very low complexity cost to continue supporting what's still called logback 1.x. (It's unfortunate that logback didn't reflect changes to their public API by bumping up the major version number.)

---

hazendaz:
Added try catch to support legacy logback 1.0.x and lower.  Added XXX to mark location since I used same method and cannot otherwise flag it deprecated support.

PR #626, opened 2016-01-26
#613 Replace Commons Logging usage with Slf4j and review logging output

diogosantana:
This is functional on tomcat 7 and 8.
I did not test it with the others application containers discussed in https://github.com/psi-probe/psi-probe/issues/613#issuecomment-174347172.

This may be breaking something, bit I guess it could be a starting point to replace commons logging.

PR #640, opened 2016-02-03
Switch logging to slf4j facet usage only and two minor spelling issues

hazendaz:
This is mostly a subset of #626 but does not introduce additional unused loggers or address some output messages that are helpful.  I've asked @diogosantana to rebase once this comes in so we can get the remainder of changes being suggested.

### DIFF
--- a/core/src/main/java/psiprobe/AbstractTomcatContainer.java
+++ b/core/src/main/java/psiprobe/AbstractTomcatContainer.java
@@ -17,8 +17,6 @@ import org.apache.catalina.Host;
 import org.apache.catalina.Valve;
 import org.apache.catalina.Wrapper;
 import org.apache.catalina.core.StandardContext;
-import org.apache.commons.logging.Log;
-import org.apache.commons.logging.LogFactory;
 import org.apache.commons.modeler.Registry;
 import org.apache.jasper.EmbeddedServletOptions;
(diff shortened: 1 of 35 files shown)


MALfunction84:
Can we also remove the Commons Logging dependencies?

hazendaz:
It was previously removed.  What it has is jcl-over-slf4j which is tricked to act like commons-logging.  We still need that for libraries like spring that log to commons-logging so we don't lose logs.  So in that case what happens is that spring continues thinking it's logging to commons-logging which jcl-over-slf4j intercepts and then sends to slf4j and ultimately to whatever logger we are using (logback in this case).

diogosantana:
I can't rebase without too much conflicts due to package renamed. You could apply this so I can make the others changes.

hazendaz:
Sorry about that.  I'll merge the code tonight.  My changes do conflict some based on your logic for logging.  We can talk more about that in issues if you want to open one to discuss it further.  Do go ahead on adding loggers to prevent us from using tomcat though as that makes sense now.


Sent from Outlook Mobile






On Wed, Feb 3, 2016 at 11:54 PM -0800, "diogosantana" <notifications@github.com> wrote:





I can't rebase without too much conflicts due to package renamed. You could apply this so I can make the others changes.

---

PR #641, opened 2016-02-03
Add SIFT logback.xml for support of testing our SIFT logback logic

hazendaz:
Please note that as noted in readme this is for manual usage to verify our SIFT appender logic with logback so we don't need to look elsewhere.  As I didn't see a real good place to put this, I just added it to the resources package on test side for web module.  If there might be a more appropriate location to store this, please advice and I'll move it.

### DIFF
--- a/web/src/test/java/resources/README.md
+++ b/web/src/test/java/resources/README.md
@@ -1,0 +1,3 @@
+Misc. Tests for verification purposes.
+
+- logback.xml has SIFT appender example in it specifically for testing SIFT support with logback.  This is not used with a test directly but rather needs bundled with app and ran in container.
--- a/web/src/test/java/resources/logback.xml
+++ b/web/src/test/java/resources/logback.xml
@@ -1,0 +1,43 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE project>
+<configuration scan="true" debug="false">
+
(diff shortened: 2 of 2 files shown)


MALfunction84:
I'm not sure I understand the purpose of this. Did you mean to include some tests with it?

hazendaz:
When I made the SIFT fix, I modified our actual logback.xml file then used that to build psi-probe to test out the solution.  It doesn't make sense for us to have that in our main logging but I didn't want to lose what I found when we need to verify SIFT appender logic.  This works by overlaying ours so that SIFT shows up without errors on the log tab.  Just wasn't sure exactly where we might keep something like this.  It's out of scope in our unit tests I think.  It's more for manual usage without having to search online to figure out SIFT appender setup.

hazendaz:
I think I have a better idea on this.  Go ahead and decline the PR.  I'll start working on building out a new series of psi-probe - demos instead which will cover all our various log support for starters including this one.  That way we will have a better way to test our internal support.  Basically each demo will be simple servlet using different logging setups.  Sound like a better approach?


Sent from Outlook Mobile






On Tue, Feb 2, 2016 at 9:34 PM -0800, "Mark Lewis" <notifications@github.com> wrote:





I'm not sure I understand the purpose of this. Did you mean to include some tests with it?

---

hazendaz:
Closing this pull request.  Will create some demo applications that include such things instead.

PR #642, opened 2016-02-04
Fixes #573 Error stopping/starting an app with Probe 2.4 under Tomcat 8

### LINKED ISSUE #573: Issue when stopping/starting an application with Probe 2.4 under Tomcat 8
With tomcat 8 (8.0.24, but I doubt it comes from the subversion ?), and Probe 2.4 official release, you can't stop/start an application by clicking on the green or red status in the application list.

It generates an error on logs
ERROR com.googlecode.psiprobe.controllers.apps.AjaxToggleContextController  - java.lang.NoSuchMethodError: org.apache.catalina.Context.getAvailable()Z

You can still stop/start it by clicking on the application and then using the stop/start button on top though as a workaround, but it was nice to be able to do it from the main screen.

It works fine with Tomcat 7 FYI.

diogosantana:
org.apache.catalina.Context.getAvailable() were deprecated in Tomcat 7
and a note were added stating it will be removed in Tomcat 8. Therefore,
the java.lang.NoSuchMethodError at
org.apache.catalina.Context.getAvailable().

### DIFF
--- a/core/src/main/java/psiprobe/controllers/apps/AjaxToggleContextController.java
+++ b/core/src/main/java/psiprobe/controllers/apps/AjaxToggleContextController.java
@@ -33,7 +33,7 @@ public class AjaxToggleContextController extends ContextHandlerController {
 
     if (!request.getContextPath().equals(contextName) && context != null) {
       try {
-        if (context.getAvailable()) {
+        if (context.getState().isAvailable()) {
           logger.info(request.getRemoteAddr() + " requested STOP of " + contextName);
           getContainerWrapper().getTomcatContainer().stop(contextName);
         } else {


diogosantana:
The test is failed test is due WhoisTests timeout. Should we continue with this unit test that depends on a external resource? This is a anti-pattern...

hazendaz:
I'll take a quick look later tonight.


Sent from Outlook Mobile






On Wed, Feb 3, 2016 at 11:46 PM -0800, "diogosantana" <notifications@github.com> wrote:





The test is failed test is due WhoisTests timeout. Should we continue with this unit test that depends on a external resource? This is a anti-pattern...

---

PR #645, opened 2016-02-04
FollowController always open log files with system charset encoding

### LINKED ISSUE #644: FollowController always open log files with default system charset encoding
FollowController always open log files with default system charset encoding even if the respective logger uses a different one.

For example, if the system default charset encoding is ISO-8859-1 and a deployed app has logback configured to UTF-8, psi-probe will not show the accented characters correctly.

diogosantana:
Fixes #644.

This actually will only work for logback loggers. We could create one issue for each logging API.

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -169,7 +169,7 @@
 		<dependency>
 			<groupId>ch.qos.logback</groupId>
 			<artifactId>logback-classic</artifactId>
-			<scope>runtime</scope>
+			<scope>compile</scope>
 		</dependency>
 		<dependency>
 			<groupId>org.quartz-scheduler</groupId>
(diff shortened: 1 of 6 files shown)


hazendaz:
I'll defer this to rest of the team to take a look at.  While this does no harm with logback it highlights a potential issue if we were to address this with other loggers, I think.  That is changing the requirement that loggers are in compile scope which could be problematic.  I'm ok with this one in general as it is otherwise what we switched to and used internally but I do want to get feedback if the approach is correct.  Generally it seems there is a lot of reflection used otherwise to prevent us from putting libraries into compile/runtime scope (logback excluded).  Maybe it doesn't really matter as we just will include additional facets.  Not 100% sure.  Thoughts?

diogosantana:
Yeah, I changed it because we do switch to logback and it might be the only one we can do that. The others we should continue to do reflections. We could easily change it back if we need to.

hazendaz:
Can you open a separate issue?  I'll merge this one :+1:

PR #649, opened 2016-02-06
Option to download compressed log files - Fixes #57

### LINKED ISSUE #57: Compress downloaded log files before sending
```
We can compress log file if an user want to download big log file. 
```

Original issue reported on code.google.com by `asp...@gmail.com` on 8 Jan 2010 at 12:48

diogosantana:
I added a img to download the log file in zip format.

Note that I don't create a temp zip file because it could take too long to start the download. The downside is we can't say the file size in advance.


hazendaz:
Something is up with your line endings on xml and property files.  I've seen GIT act up like this when using Eclipse EGIT to commit files.  Not sure if that is how you are doing it.  I pulled this PR locally and just pushed it with a fix.  I'm going to merge it that way.

diogosantana:
Ok, thanks. I do use EGIT and it doesn't show me those changes in line endings.

I will try to figure it out next time.

PR #654, opened 2016-02-06
Move jcl-over-slf4j into parent and update to 1.7.14

hazendaz:


hazendaz:
Movement of this was to ensure we catch the version upgrade much more easily as the slf4j-api was already at 1.7.14 but the core jcl-over-log4j was at 1.7.13.

PR #655, opened 2016-02-06
Cleanup exception handling per sonar results

hazendaz:
This covers majority of exceptions but still leaves quite a few caught Exception where there was more than one line of additonally processing in the exception.  For those, I sort of felt rather than duplicating code we can wait until we are on java 7 and can use multi-catch.

### DIFF
--- a/core/src/main/java/psiprobe/beans/ContainerWrapperBean.java
+++ b/core/src/main/java/psiprobe/beans/ContainerWrapperBean.java
@@ -125,9 +125,7 @@ public class ContainerWrapperBean {
         logger.info("Unregistering container adapter");
         tomcatContainer.setWrapper(null);
       }
-    } catch (ThreadDeath e) {
-      throw e;
-    } catch (Throwable e) {
+    } catch (Exception e) {
       logger.error("Could not unregister container adapter", e);
     }
   }
--- a/core/src/main/java/psiprobe/tools/AccessorFactory.java
+++ b/core/src/main/java/psiprobe/tools/AccessorFactory.java
@@ -43,7 +43,13 @@ public class AccessorFactory {
   private static Accessor getReflective() {
     try {
       return new ReflectiveAccessor();
-    } catch (Exception ex) {
+    } catch (ClassNotFoundException e) {
+      return null;
+    } catch (InstantiationException e) {
+      return null;
+    } catch (IllegalAccessException e) {
+      return null;
+    } catch (NoSuchMethodException e) {
       return null;
     }
   }
--- a/core/src/main/java/psiprobe/tools/ReflectiveAccessor.java
+++ b/core/src/main/java/psiprobe/tools/ReflectiveAccessor.java
@@ -54,7 +54,11 @@ public class ReflectiveAccessor implements Accessor {
       if (fieldAccessor != null) {
         return get.invoke(fieldAccessor, new Object[] {obj});
       }
-    } catch (Exception ex) {
+    } catch (IllegalArgumentException e) {
+      // ignore
+    } catch (IllegalAccessException e) {
+      // ignore
+    } catch (InvocationTargetException e) {
       // ignore
     }
     return null;
--- a/core/src/main/java/psiprobe/AbstractTomcatContainer.java
+++ b/core/src/main/java/psiprobe/AbstractTomcatContainer.java
@@ -33,6 +33,7 @@ import psiprobe.model.jsp.Summary;
 
 import java.io.File;
 import java.net.URI;
+import java.net.URISyntaxException;
 import java.net.URL;
 import java.net.URLClassLoader;
 import java.util.ArrayList;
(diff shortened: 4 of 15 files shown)


core/src/main/java/psiprobe/beans/ContainerWrapperBean.java
```diff
@@ -127,7 +127,7 @@ public void setWrapper(Wrapper wrapper) {
       }
     } catch (ThreadDeath e) {
```
MALfunction84: `catch (ThreadDeath)` is only necessary if you're going to use `catch (Throwable)`. ThreadDeath extends Error, so it won't be caught by `catch (Exception)`. You can remove `catch (ThreadDeath)` entirely, and it'll be cleaner.

There are many instances of this throughout this PR. I won't note them all because that's just noisy.

hazendaz: Perfect!  I was looking at that yesterday and wasn't quite sure as I've never saw thread death before.


Sent from Outlook Mobile






On Sun, Feb 7, 2016 at 9:31 AM -0800, "Mark Lewis" <notifications@github.com> wrote:





> @@ -127,7 +127,7 @@ public void setWrapper(Wrapper wrapper) {
>        }
>      } catch (ThreadDeath e) {

`catch (ThreadDeath)` is only necessary if you're going to use `catch (Throwable)`. ThreadDeath extends Error, so it won't be caught by `catch (Exception)`. You can remove `catch (ThreadDeath)` entirely, and it'll be cleaner.

There are many instances of this throughout this PR. I won't note them all because that's just noisy.

---
Reply to this email directly or view it on GitHub:
https://github.com/psi-probe/psi-probe/pull/655/files#r52120883

hazendaz:
Rebased...thread death now removed.

core/src/main/java/psiprobe/tools/AccessorFactory.java
```diff
@@ -43,7 +43,13 @@ public static Accessor getInstance() {
   private static Accessor getReflective() {
     try {
       return new ReflectiveAccessor();
-    } catch (Exception ex) {
+    } catch (ClassNotFoundException e) {
```
MALfunction84: Since we're on minimum JDK 7, you could use multi-catch here.

core/src/main/java/psiprobe/tools/ReflectiveAccessor.java
```diff
@@ -54,7 +54,11 @@ public Object get(Object obj, Field field) {
       if (fieldAccessor != null) {
         return get.invoke(fieldAccessor, new Object[] {obj});
       }
-    } catch (Exception ex) {
+    } catch (IllegalArgumentException e) {
```
MALfunction84: Could also use multi-catch here. We'd have to update the root pom.xml though. Maybe save this for another PR.

core/src/main/java/psiprobe/tools/AccessorFactory.java
hazendaz: At the moment, we are still compiling to java 6.  Do you know if multi-catch is something we can code and compile towards 6?  We haven't bumped up fully to 7 yet.

jander99: CodePro and/or Findbugs will flag that as an issue due to lack of logging.
With the other logging work stream going we can ignore for now, but once
logging is finalized, let's come back and add logging (even if debug) in
places like this and others.

hazendaz: I've got this already...I was waiting on on Pull request which @MALfunction84 just merged.  I'll review what I have and submit full logging coverage next.  I've even switched everything to use {} on loggers to prevent early processing in case logging is turned off.

hazendaz: What is CodePro?  First I've heard of that.

jander99: https://developers.google.com/java-dev-tools/codepro/doc/?hl=en

Eclipse will complain because this hasn't been updated for 4+ yet.  The 3.7
version still works in Luna.

MALfunction84: Yeah, I noted elsewhere we'd have to bump the property values to 1.7 in the root pom.xml to use multi-catch. I don't see any problem with pulling the trigger on that now.

hazendaz: Perfect!  Java 1.7 it is.  Tomcat 7 does require that if wanting to use websockets.  I thought I asked that out on the google groups but didn't hear anything back.  I'll check again to be certain.  Once we make that jump...I see the code immediately taking full advantage so there won't be any going back.  Only issue is we will be jumping from 5 to 7 without any java 6 support at all.  Any concern there?

MALfunction84: Nah, I'm not worried about it. We probably could have dropped Java 5 support in 2.4 given its EOL in May 2015.

PR #656, opened 2016-02-07
Replaces Commons Logging usage with Slf4j - Fixes #613

### LINKED ISSUE #613: Replace Commons Logging usage with Slf4j and review logging output
As of #612, we are officially using logback logging rather than legacy log4j 1.  Now we need to address switching over our own internal logging to use slf4j so that we are not wrapped through jcl-to-slf4j needlessly in order to improve performance.  Additionally, this should be an opportunately to review logging configuration in case any adjustments are necessary.  The basic configuration as it stands is similar to that of tomcat logging at this time.

diogosantana:
Some loggers atributes were added in classes that don't use it. The
reason is to prevent usage of Spring super classes protected logger
atribute.

@hazendaz can you check if the line endings problem is occurring?

### DIFF
--- a/core/src/main/java/psiprobe/AbstractTomcatContainer.java
+++ b/core/src/main/java/psiprobe/AbstractTomcatContainer.java
@@ -406,7 +406,7 @@ public abstract class AbstractTomcatContainer implements TomcatContainer {
   }
 
   @Override
-  public org.apache.juli.logging.Log getLogger(Context context) {
+  public Object getLogger(Context context) {
     return context.getLogger();
   }
 
--- a/core/src/main/java/psiprobe/ProbeServlet.java
+++ b/core/src/main/java/psiprobe/ProbeServlet.java
@@ -13,6 +13,8 @@ package psiprobe;
 
 import org.apache.catalina.ContainerServlet;
 import org.apache.catalina.Wrapper;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.springframework.web.servlet.DispatcherServlet;
 
 import psiprobe.beans.ContainerWrapperBean;
@@ -34,6 +36,9 @@ public class ProbeServlet extends DispatcherServlet implements ContainerServlet 
 
   /** The Constant serialVersionUID. */
   private static final long serialVersionUID = 1L;
+
+  /** The logger. */
+  protected Logger logger = LoggerFactory.getLogger(getClass());
 
   /** The wrapper. */
   private Wrapper wrapper;
--- a/core/src/main/java/psiprobe/controllers/ContextHandlerController.java
+++ b/core/src/main/java/psiprobe/controllers/ContextHandlerController.java
@@ -12,6 +12,8 @@
 package psiprobe.controllers;
 
 import org.apache.catalina.Context;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.springframework.web.bind.ServletRequestUtils;
 import org.springframework.web.servlet.ModelAndView;
 
@@ -25,6 +27,9 @@ import javax.servlet.http.HttpServletResponse;
  * @author Mark Lewis
  */
 public abstract class ContextHandlerController extends TomcatContainerController {
+
+  /** The logger. */
+  protected Logger logger = LoggerFactory.getLogger(getClass());
 
   @Override
   protected ModelAndView handleRequestInternal(HttpServletRequest request,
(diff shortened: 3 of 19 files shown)


hazendaz:
Should be fine.  So far I've only seen line ending issues on new files.


Sent from Outlook Mobile






On Sun, Feb 7, 2016 at 1:41 AM -0800, "diogosantana" <notifications@github.com> wrote:





Some loggers atributes were added in classes that don't use it. The
reason is to prevent usage of Spring super classes protected logger
atribute.

@hazendaz can you check if the line endings problem is occurring?
You can view, comment on, or merge this pull request online at:

  https://github.com/psi-probe/psi-probe/pull/656

-- Commit Summary --

  * Replaces Commons Logging usage with Slf4j - Fixes #613

-- File Changes --

    M core/src/main/java/psiprobe/AbstractTomcatContainer.java (2)
    M core/src/main/java/psiprobe/ProbeServlet.java (5)
    M core/src/main/java/psiprobe/controllers/ContextHandlerController.java (5)
    M core/src/main/java/psiprobe/controllers/RenderChartController.java (5)
    M core/src/main/java/psiprobe/controllers/TomcatContainerController.java (5)
    M core/src/main/java/psiprobe/controllers/WhoisController.java (5)
    M core/src/main/java/psiprobe/controllers/apps/AjaxReloadContextController.java (2)
    M core/src/main/java/psiprobe/controllers/apps/AjaxToggleContextController.java (2)
    M core/src/main/java/psiprobe/controllers/apps/NoSelfContextHandlerController.java (2)
    M core/src/main/java/psiprobe/controllers/deploy/UndeployContextController.java (2)
    M core/src/main/java/psiprobe/controllers/deploy/UploadWarController.java (2)
    M core/src/main/java/psiprobe/controllers/sql/CachedRecordSetController.java (5)
    M core/src/main/java/psiprobe/controllers/sql/QueryHistoryItemController.java (5)
    M core/src/main/java/psiprobe/controllers/system/AdviseGarbageCollectionController.java (5)
    M core/src/main/java/psiprobe/controllers/threads/GetClassLoaderUrlsController.java (5)
    M core/src/main/java/psiprobe/controllers/wrapper/RestartJvmController.java (5)
    M core/src/main/java/psiprobe/controllers/wrapper/StopJvmController.java (5)
    M core/src/main/java/psiprobe/controllers/wrapper/ThreadDumpController.java (5)
    M core/src/main/java/psiprobe/controllers/wrapper/WrapperInfoController.java (5)

-- Patch Links --

https://github.com/psi-probe/psi-probe/pull/656.patch
https://github.com/psi-probe/psi-probe/pull/656.diff

---
Reply to this email directly or view it on GitHub:
https://github.com/psi-probe/psi-probe/pull/656

core/src/main/java/psiprobe/AbstractTomcatContainer.java
```diff
@@ -406,7 +406,7 @@ public void listContextJsps(Context context, Summary summary, boolean compile) {
   }
 
   @Override
-  public org.apache.juli.logging.Log getLogger(Context context) {
+  public Object getLogger(Context context) {
```
MALfunction84: What motivated this change? All supported Tomcat versions return `org.apache.juli.logging.Log`.

hazendaz: Main objective is to make sure we use our own loggers and don't mistakening use tomcats.  In this case, not once is this method ever used other than as Object.  However, I see zero point for this in the interface or the concrete class.  It's 3 use cases simply call through the wrapper very pointlessly passing 'context' which then just reads context.getLogger().  The wrapper calls provide no additional processing and in my opinion are not needed.  Instead we should delete the interface method, concrete method in abstract class, and then just call it directly thus not wasting time calling through objects.

After deleting those, in LogResolverBean, the 3 calls all will be the same.

Previous
```
Object contextLogger = getContainerWrapper().getTomcatContainer().getLogger(ctx);
```
New would still just receive object back because it can be considered different in a couple of the calls plus setting wants Object not actual log type.
```
Object contextLogger = ctx.getLogger();
```
Please take a quick look over this to see if I'm missing anything in this actual use case.  I've got this switched locally as my thinking suggests above.  I'll be running some tests shortly.

MALfunction84: Most likely a holdover from an older version of Tomcat. What you're saying makes sense. Let's remove it.

hazendaz: Is it ok to do that on a separate Pull Request?  I don't see much harm letthing this one through then following up with that fix.

hazendaz:
OK - no further comment.  I'm merging as fix will come later to remove the unnecessary method and I'm trying to apply proper usage of slf4j without string concat.  There are a few incorrect here so I'll pick those up on my branch.  Loggers should be static final but that can be addressed later.  The main objective here is completed.  Sonar will notify us on the rest :)

PR #674, opened 2016-02-15
Create new maven module, add Spring HATEOAS package.

jander99:
Here is a PR that adds a new module to PSI Probe, rest.  This new rest module will contain the logic required to build RESTful services of the data required by JSPs.  Since we have not come up with a layout of base URLs for the RESTful endpoints, this module is currently empty.  I will create several more PRs leading up to a working status RESTful service.  This will be the blueprint from which all other RESTful services will be created.

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -156,8 +156,8 @@
 			<scope>compile</scope>
 		</dependency>
 		<dependency>
-			<groupId>org.jboss.spec.javax.transaction</groupId>
-			<artifactId>jboss-transaction-api_1.2_spec</artifactId>
+			<groupId>javax.transaction</groupId>
+			<artifactId>javax.transaction-api</artifactId>
 			<scope>compile</scope>
 		</dependency>
 		<dependency>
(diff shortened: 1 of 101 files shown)


hazendaz:
@jander99 Due to the restrictions I set in github, it requires the PRs to be rebased before submission otherwise we cannot merge.  We may want to disable that requirement but for now, I think it is managable.  Just requires a quick rebase and we need to do so on each one as we sequentially bring them in.

jander99:
Can I give up now? :)

hazendaz:
I'm going to look at the github security settings again.  Obviously this is more of a pain than help.


Sent from Outlook Mobile






On Mon, Feb 15, 2016 at 12:47 PM -0800, "Jeff Anderson" <notifications@github.com> wrote:





Can I give up now? :)

---

jander99:
Nah this is my fault not knowing git very well.  I'm learning.  Looks like the new PR is stuck on travis but travis doesnt know anything about the PR.

hazendaz:
I'll leave the github settings be for now.  I looked and it doens't really call out what it is forcing on us.  However, what it does protect seems woth keeping.  This also does ensure code comes in to the master in sequence.  It's too bad though they don't allow us to choose on that setting as it otherwise causes us to rebase a lot if we are all working at same pace.  I think overall it's managable and as you stated, you don't know git very well.  That will certainly change with these rules in place :)  If it does end up being too much of a bother though we can reduce the enforcement.

Issue #659, opened 2016-02-08
Show TrustStore certificates


sandrapascual:
The issue is about showing a list with the certificates contained in TrustStore defined in the System Property "javax.net.ssl.trustStore" sorted by Expiration Date.

This new funcionality will be shown in the system tab down to Wrapper Control.

farconada:
It's a great idea. I need something like this to find the root cause of ssl handshake exceptions in fast way

hazendaz:
+1


Sent from Outlook Mobile






On Mon, Feb 8, 2016 at 5:46 AM -0800, "Fernando Arconada" <notifications@github.com> wrote:





It's a great idea. I need something like this to find the root cause of ssl handshake exceptions in fast way

---

hazendaz:
@farconada ignore previous comment if it reached you.  Posted in wrong spot.

diogosantana:
I can try to do this as I have a lot of xp with certificates, jks and SSL/TLS, although it won't be enough to debug ssl handshake exceptions.

jander99:
I think the point would be to see whether a keystore has the necessarily
certificates.  For instance, where I work, we have our own internal SSL
CA.  If the JDK is not patched to include the CA's cert, the handshake
fails.  It would be nice to be able to see the list (output from keytool,
basically) of certs and determine if a JDK has the necessarily certs.

diogosantana:
I noticed https://github.com/sandrapascual/psi-probe/commit/081ab4039f71276c7c86f877e21980cdd94ee615 commit only list the jvm configured trust stores (`javax.net.ssl.trustStore`) certificates, but tomcat accepts configuration of two stores per connector, the server certificate store and the trusted certificates store.

diogosantana:
Oh, I just saw the PR conversation... nvm

jander99:
Just to document, you mean the keystoreFile and truststoreFile from the
Connector documentation
https://tomcat.apache.org/tomcat-8.0-doc/config/http.html#SSL_Support ?

diogosantana:
Yes. But I do agre with @sandrapascual PR's comment (https://github.com/psi-probe/psi-probe/pull/660#issuecomment-181757598) about put this on system tab and later add per connector config.

Maybe we should add another tab like "Certificates" with all system and per connector configs. Thoughts?

hazendaz:
[...]

Issue #596, opened 2016-01-19
Test Cases and mocking


hazendaz:
Current test coverage on this library is incredibly low at 5 to 6 percent.  Need to raise that which probably means we will need some mocking framework as well.  While this framework I'm going to suggest is not for beginners, after over a year with heavy usage, it blows away the older mocking libraries.  So I'd like to see test coverage be increased to at least 50% by end of 2016 and for mocking to be performed via JMockit framework.  The higher the precentage we can get, the lower the sonar scores will go.  Currently sonar expects 65% per class.

hazendaz:
#606 increases code coverage slighly and improves container determination
#607 increases code coverage considerbly but does not test more complicated logic.  It only cares that javabeans are properly setup.  This did discover a few constructor issues.  It further found some issues with constructors I was not fully able to understand so those tests are @Ignore for time being.

Coverage as of these changes is around 25%.

hazendaz:
Closing as separate tickets opened for milestone 30%, 40%, etc.  We are actually at 29.x% now.

PR #676, opened 2016-02-16
Created new tab Certificates

diogosantana:
Created new tab Certificates as a most complete solution to #659.

The tab will show the following list of certificates:

- System key store certificates
- System trust store certificates
- Connectors key store certificates
- Connectors trust store certificates

@hazendaz please let me know if it still is with line ending problem


hazendaz:
@diogosantana Line endings problems have been resolved.  Thanks.

As for this change, what should I be expecting to occur?  I've got certificates setup for tomcat only on the connectors.  In my specific use-case on my local machine I just have a keystore.  When I pull this PR and run a quick test, it's essentially other than showing the following words...

Connectors Certificates
http-nio-8443

What I don't have setup are JVM arguments and I sort of feel those are a problem in general as it allows the password to easily be determined.  My assumption was that this PR might show me results similar to that if I had the JVM properties defined.  Is that not yet the case?

diogosantana:
Hum... I test it all and work on a Tomcat 8.0.32 (the zipped one).
Any log errors?

diogosantana:
This is my example with two connectors.
probe-certificates

hazendaz:
System Certificates on the certificate tab work just fine.  Connectors one is not working.  It might be at least helpful to review the code a little more to post some sort of message like "not configured" in case there are none.  It's possible my configuration is not 100% correct.  Until last week, I thought it was ok to have both keystore and truststore in same file.  I know a bit more now.  So potentially I have something else setup wrong.  I'm otherwise ok with the changes.

@jander99 
@MALfunction84 

Can either of you give this a test and see if you are able to accurately see the connector certificates?  I'm very eager for this to work :)  Ultimately if this works, I think next is ability to add/delete/update/etc is quite possible and saves a huge amount of work for a system admin that doesn't do this often :)

hazendaz:
@diogosantana 

Now that you are bringing jmockit into the project (+1)...maybe you can work out some more tests to raise general coverage in other areas.  I'm all for jmockit even though I personally feel the owner of jmockit needs a more consistent api from version to version but at least 1.21 and few versions before are extremely stable for anyone with past experience with jmockit library.  It used to be a really rocky ride from version to version.  I've been using it for a couple of years now over roughly 100K lines of code so I know it's power and it's power is awesome outside of my complaints!

diogosantana:
@hazendaz Can you show your connector config? By the way, this should only show JSSE. It won't show connectors that are not marker as SECURE or that uses APR/Native or AJP.

hazendaz:
Here is the general setup I use...

<Connector SSLEnabled="true" clientAuth="false" keystoreFile="someKeystoreFile.jks" keystorePass="xxxxxxxx" port="8443" scheme="https" secure="true" sslProtocol="TLS"/>

Does that look correct?

diogosantana:
@hazendaz About jmockit, I added it because I was needing mock framework to test properly and as a result of "discussion" on #596. But I'm used to use Mockito, so expect some wrongness...

My view is we don't need to spend time creating test for things we are not touching. As we create more stuff or change others, then we create tests for those.

hazendaz:
On jmockit...no worries.  I was hoping you were very familiar with it to spread it through the code base quickly.  It will get there eventually :)

diogosantana:
The config seams to be ok. Try to add ` protocol="org.apache.coyote.http11.Http11NioProtocol"`.
No errors on log? Can you debug it?

hazendaz:
I have that setting too although that is default in tomcat 8 (BIO was deprecated).  I have a number of other settings too that I felt were unrelated.  I'll try debugging it.

hazendaz:
I found the issue.  Current code doesn't support catalina.base or catalina.home starting path of keystore/truststore as allowed by tomcat.  I adjusted the code a bit on my local copy to use catalina.home.  That worked as I expected.  I assume there is a better tomcat way to read the file location and/or it's going to need to look as you have it plus catalina.base or catalina.home.

In toConnectorInfo...

if file doesn't exist, I added this to test my use-case which I think is pretty normal....

```java
 else {
        file = new File(System.getProperty("catalina.home") + File.separator + keystoreFile);
        if (file.exists()) {
          info.setKeyStoreFile(file);
        }
      }
```

diogosantana:
It makes sense. I will look at tomcat source to be 100% sure.

Btw, I always put the full path to avoid problem... thats why it work fine
for me.
Em 16/02/2016 02:16, "Jeremy Landis" <notifications@github.com> escreveu:

diogosantana:
There is no "System Certificates" as I put it on the view. Configuration is always per connector.

Also Tomcat never looks at `javax.net.ssl.keyStore`, but it looks at `javax.net.ssl.trustStore` if no `truststoreFile` is provided by connector configuration. So I will reproduce the same behavior to list certificates per secure connector configured.

diogosantana:
@hazendaz 

Btw, Tomcat source code only looks at catalina.base, not catalina.home as you say it worked in your environment. Can you confirm that?

hazendaz:
I have both set as the same location..ie default tomcat setup.  So if that is what tomcat docs state I would agree.


Sent from Outlook Mobile






On Wed, Feb 17, 2016 at 5:43 PM -0800, "diogosantana" <notifications@github.com> wrote:





@hazendaz

Btw, Tomcat source code only looks at catalina.base, not catalina.home as you say it worked in your environment. Can you confirm that?

---

hazendaz:
@diogosantana I'm going pull this in on separate PR.  I have squashed the code leaving only the jar introductions separate.  I also found a bug in cert handling as it did not handle null pointers when cert information is not available on connector.  I did remove the one closeable that was added since it was later not used as a result of java 7.  And finally one minor log cleanup.

This is working for me now.  Thank you so much for all your hard work!  This is really great!

hazendaz:
Closed via #678

PR #680, opened 2016-02-21
Build cleanup - no impact

hazendaz:
Small build cleanup with no impact.


hazendaz:
This requires no review as it is related to build only and clean build is the test.

PR #684, opened 2016-02-28
Update versions per versioneyes

hazendaz:


hazendaz:
Low impact change which has been out for a number of days.  I'm going to go ahead and merge.  No source, just dependency updates and site page updates.

Issue #687, opened 2016-03-04
Connectors page only showing "psi-probe-web"


e1d:
I downloaded and deployed the latest snapshot (Version 3.0.0-SNAPSHOT).

Everything works fine, except on the Connectors page, it doesn't show for example "ajp-bio-8009" connector, or "http-nio-8080". It only showed "psi-probe-web".

hazendaz:
Can you post a screen shot?  Also can you try milestone 3 and let me know if the same problem exists there?  Currently milestone 3 is ahead of snapshot.  Thanks.


Sent from Outlook Mobile






On Fri, Mar 4, 2016 at 12:17 AM -0800, "e1d" <notifications@github.com> wrote:





I downloaded and deployed the latest snapshot (Version 3.0.0-SNAPSHOT).

Everything works fine, except on the Connectors page, it doesn't show for example "ajp-bio-8009" connector, or "http-nio-8080". It only showed "psi-probe-web".

---
Reply to this email directly or view it on GitHub:
https://github.com/psi-probe/psi-probe/issues/687

hazendaz:
Also what version of tomcat are you using?


Sent from Outlook Mobile






On Fri, Mar 4, 2016 at 12:17 AM -0800, "e1d" <notifications@github.com> wrote:





I downloaded and deployed the latest snapshot (Version 3.0.0-SNAPSHOT).

Everything works fine, except on the Connectors page, it doesn't show for example "ajp-bio-8009" connector, or "http-nio-8080". It only showed "psi-probe-web".

---
Reply to this email directly or view it on GitHub:
https://github.com/psi-probe/psi-probe/issues/687

e1d:
I'm using tomcat 8 & 7. Both returning same result.
Here is my screenshot 
![connector display](https://cloud.githubusercontent.com/assets/10696061/13558784/9097e770-e445-11e5-88c0-7df5ab441871.jpg)

e1d:
Also, I got this error when creating war file with maven. What could be the problem?

`[INFO]
[INFO] --- git-commit-id-plugin:2.1.13:revision (git-commit-id) @ core ---
[INFO] ------------------------------------------------------------------------
[INFO] Reactor Summary:
[INFO]
[INFO] PSI Probe .......................................... SUCCESS [ 27.791 s]
[INFO] PSI Probe Core ..................................... FAILURE [04:55 min]
[INFO] PSI Probe Tomcat 5.0.x Adaptor ..................... SKIPPED
[...]

PR #688, opened 2016-03-06
Qualify all field/method access

hazendaz:
As part of an effort to apply checks that eclipse can flag, qualify all access which additionally makes it easier to understand by look at code where object in use comes from.

### DIFF
--- a/core/src/main/java/psiprobe/AbstractTomcatContainer.java
+++ b/core/src/main/java/psiprobe/AbstractTomcatContainer.java
@@ -82,28 +82,28 @@ public abstract class AbstractTomcatContainer implements TomcatContainer {
   public void setWrapper(Wrapper wrapper) {
     Valve valve = createValve();
     if (wrapper != null) {
-      host = (Host) wrapper.getParent().getParent();
-      Engine engine = (Engine) host.getParent();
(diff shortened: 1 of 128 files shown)


hazendaz:
@jander99 @MALfunction84 

If  possible can I get an ok on this so I can then bring in a major fix around issue #687 which is the long standing issue I have been mentioning where grids didn't look right.  Found the issue.  I indirectly introduced it back in July of last year.  I'll be making a pull request on 2.4.x branch as well to fix the problem and plan to release 2.4.1.

MALfunction84:
I generally don't like large changes, and I think this specific change is not necessary. Usage of `this` that is not required makes the code busier and harder to read.

If a local variable hides a field, you should just rename one of them. Many IDEs provide a warning for this. (I actually have my syntax highlighting set up to color fields and local variables differently.)

My one exception to this rule is for field assignments:

    class Example {
      Object value;

      Example(Object value) {
        this.value = value; // OK
      }   

      void setValue(Object value) {
        this.value = value; // OK
      }

      void doSomething() {
        Object value = new Object(); // not OK, hides field
      }
    }

In short, I don't like this change. Can you create a separate PR for fixing #687?

PR #690, opened 2016-03-08
Fix defect introduced by #527 causing maven to filter connctions.jsp …

hazendaz:
…resulting in use of project name

### DIFF
--- a/core/src/main/java/com/googlecode/psiprobe/beans/ContainerListenerBean.java
+++ b/core/src/main/java/com/googlecode/psiprobe/beans/ContainerListenerBean.java
@@ -250,7 +250,7 @@ public class ContainerListenerBean implements NotificationListener {
         ObjectName poolName = threadPoolObjectName.getThreadPoolName();
 
         Connector connector = new Connector();
-        connector.setName(poolName.getKeyProperty("name"));
+        connector.setProtocolHandler(poolName.getKeyProperty("name"));
 
         ObjectName grpName = threadPoolObjectName.getGlobalRequestProcessorName();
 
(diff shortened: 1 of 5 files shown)


hazendaz:
Fixes #687.  This issue has been long standing affecting 2.4.0 release.  It was noted on each 3.0.0 milestone and with help from @e1d this has been resolved.

MALfunction84:
Rather than adjusting the filtering, can we change the name of the property? Or if we're going to adjust filtering, it would be more consistent/predictable to filter out entire directories rather than single files.

hazendaz:
Long term that was the plan.  This was more of a quick fix.  I'll look at changing the name of the field tonight and will apply to both 2.4 and 3.0.  I'll also remove my other PR.


Sent from Outlook Mobile






On Mon, Mar 7, 2016 at 7:57 PM -0800, "Mark Lewis" <notifications@github.com> wrote:





Rather than adjusting the filtering, can we change the name of the property? Or if we're going to adjust filtering, it would be more consistent/predictable to filter out entire directories rather than single files.

---

hazendaz:
@MALfunction84 The name has been changed to protocolHandler.  As you will see it was quite a bit more extensive to pull off.  It did take me a few tries to catch everything so an extra set of eyes in there would be much appreciated.

MALfunction84:
:+1:
@hazendaz When you backport this to 2.4, please change the milestone of #687 to 2.4.1 for changelogs and such. If you have any other fix backport candidates in mind for 2.4.1, leave a note on the issue and mention me.

PR #691, opened 2016-03-08
Removed unnecessary small font sizes.

nning:


web/src/main/webapp/css/classic/mainnav.css
```diff
@@ -26,7 +26,6 @@
 #tabnav a:link, #tabnav a:visited {
 	float: left;
 	background: #f3f3f3;
-	font-size: 12px;
```
hazendaz: Where exactly do I see this change on the page?

web/src/main/webapp/css/classic/main.css
```diff
@@ -11,7 +11,6 @@
 
 BODY {
 	font-family: "Trebuchet MS", Arial, Helvetica, sans-serif;
-	font-size: 75%;
```
hazendaz: @MALfunction84 Are you ok with this L&F change?

psi-probe

MALfunction84: Heh, I'm so used to the small fonts, that just looks gigantic.

I'm a bit concerned how it will look on narrower monitors. I think we're well past the era of 1024x768, but 1440w isn't exactly embarrassing. How does it look in a smaller window?

The vertical alignment of icons looks off. If people think the interface needs work, we should consider an overhaul, maybe using Bootstrap or something.

nning: I just had problems reading anything on a 15" display with 1080p resolution...

### 1440x960

2016-03-09-082250_5760x1080_scrot

### 1024x768

2016-03-09-082306_5760x1080_scrot


(:+1: for an additional, responsive interface overhaul!)

MALfunction84: That actually looks pretty good on lower resolutions. Still, the icon vertical alignment is an issue.

nning: Do you mean the icons inside the table?

I googled `psi probe screenshot` and this screenshot indicates the vertical icon aligment is slightly off to the top since version 2.0.2 (at least):
http://i1-scripts.softpedia-static.com/screenshots/PSI-Probe_1.png

PR #694, opened 2016-03-13
Correct SCM Urls

hazendaz:
Had a ':' where '/' expected.


hazendaz:
This was discovered when trying to setup another project to follow similar settings and then releasing using maven plugin.  This was yet to be seen as issue here but would as soon as mvn release:* is used.  Merging...

PR #700, opened 2016-03-19
Update to commons-lang3 3.4

hazendaz:


hazendaz:
Due to issue in javabean-tester associated with Dates after most recent changes that occurs randomly, we experience build errors due to test failures.  Quick fix was to just skip two Date Fields that cause the condition.  It's not working as expected and thus ok now.  Will address core issue in javabean-tester in near future.

hazendaz:
Will reopen and cleanly pull in fixes so they are not mixed.  This push was to vet fix...

PR #722, opened 2016-05-01
[jsp] Cleanup syntax errors with tag usage

hazendaz:


hazendaz:
Routine change to get rid of errors within eclipse.  Just single quotes around the EL.

Issue #241, opened 2015-03-16
Allow users to remotely browse and edit application files


GoogleCodeExporter:
```
Please describe the feature you would like to see added.

I think it would be great if psi-probe could show and change the applications 
configuration files like *.xml/*.properties of each aplication, and then reload 
it

How will this feature make Probe better?

the monitoring and manteinance  of apps  could improve a lot, because we would 
be able to watch all  of our applications deployed configurations in all of the 
environments, then you would be able to change it and reload the app, making 
easier the change between enviroments, without the tedious work of change it 
manually in each environment
```

Original issue reported on code.google.com by `jorgeremar@gmail.com` on 24 Aug 2011 at 10:12

PR #723, opened 2016-05-02
Issue #241. Allow users to copy a single file into a deployed context.

sandrapascual:
Issue #241. Allow users to copy a single file into a deployed context.
Usefull to hot change report, jsp, xml and properties files.

### DIFF
--- a/core/src/main/java/psiprobe/controllers/deploy/CopySingleFileController.java
+++ b/core/src/main/java/psiprobe/controllers/deploy/CopySingleFileController.java
@@ -12,10 +12,10 @@
 package psiprobe.controllers.deploy;
 
 import java.io.File;
+import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.Vector;
 
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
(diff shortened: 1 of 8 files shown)


core/src/main/java/psiprobe/controllers/deploy/CopySingleFileController.java
```diff
@@ -0,0 +1,185 @@
...
+
+	  List<Context> apps;
+	    try {
+	      apps = getContainerWrapper().getTomcatContainer().findContexts();
+	    } catch (NullPointerException ex) {
+	      throw new IllegalStateException("No container found for your server: "
+	          + getServletContext().getServerInfo(), ex);
+	    }
+	  
+	    Vector applications = new Vector();
```
hazendaz: Can you use something other than vector here?  Vector is a bad idea for new code.

core/src/main/java/psiprobe/controllers/deploy/CopySingleFileController.java
```diff
@@ -0,0 +1,185 @@
...
+	      if (appContext.getName() != null && appContext.getName().trim().length()>0) {
+	    	  Map<String, String> app = new HashMap<String, String>();
+	    	  app.put("value", appContext.getName());
+	    	  app.put("label", appContext.getName());
+	    	  applications.add(app);
+	      }
+	    }
+	    request.setAttribute("apps", applications);
+
+	  if (FileUploadBase.isMultipartContent(new ServletRequestContext(request))) {
```
hazendaz: Please format the this class.

hazendaz:
@sandrapascual Just a couple of comments.  Otherwise looks promising :)  

Let me toss another idea out here.  This currently just addresses a deployed context, given we have pretty much full control of tomcat, and I assume this doesn't allow this, would it be possible to maybe make this even more flexible so that we could change tomcat files in bin/conf directories?  I see a need for that where password might exist we need to change rather than going to the box.  I know in my use case anyways we have the ability externally to start/stop tomcats but not to address box level changes like that.  Would be cool to have that, maybe in another pull request.  Not sure it's completely feasible but would be a nice feature.

hazendaz:
@sandrapascual When you address the small issues I noted, please rebase against lastest master.  We have a check that prevents merging otherwise.  Just be sure to rebase right before you are ready to push this again.  I may or may not have a few other changes today :)

sandrapascual:
Hi, I've change the Vector use and formated the class. 
Sorry, but I usually have problems to understand the rebase methods in git. I've tried to do it as better as I can. Please let me know if my pull-request is not good... 
Moving to the possibility of copying files to bin/conf folders, I think it could be dangerous for not experimented users... I could be a good option to configure correctly the permissions

hazendaz:
No worries on rebasing.  For reference, it's quite easy...'git pull --rebase upstream master'.  Then once that completelys perform a squash interactively ('git rebase -i HEAD~X where x is number of commits to squash).  Then simply follow directions provided by git.

Since github otherwise allows squashing when pulling this in, I'll do that.  Thanks for the contribution!

Issue #733, opened 2016-06-01
Trace important user actions in log file


sandrapascual:
Add tracing to probe.log with user and actions.

Only in important actions such as, deploy war, undeploy, start, stop, reload, copy files to context and discarding work directory when deploying or copying a file into a context.

hazendaz:
merged on #744

PR #734, opened 2016-06-01
Issue #733 Trace important user actions in log file

sandrapascual:
Add tracing to probe.log with user and actions.

Only in important actions such as, deploy war, undeploy, start, stop, reload, copy files to context and discarding work directory when deploying or copying a file into a context.

### DIFF
--- a/core/src/main/java/psiprobe/controllers/apps/AjaxReloadContextController.java
+++ b/core/src/main/java/psiprobe/controllers/apps/AjaxReloadContextController.java
@@ -14,6 +14,8 @@ package psiprobe.controllers.apps;
 import org.apache.catalina.Context;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.springframework.security.core.Authentication;
+import org.springframework.security.core.context.SecurityContextHolder;
 import org.springframework.web.servlet.ModelAndView;
 
 import psiprobe.controllers.ContextHandlerController;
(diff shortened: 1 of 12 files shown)


hazendaz:
@sandrapascual Can you reformat using 'spaces' instead of 'tabs'?  Also on logs, you can incur less cost by using {} within logging message followed by params.  Such as 'logger.error("Some wonderful {} and {}", 'checks', 'balances').  This is important if logging levels are manipulated so we don't incur cost to create concatenation and then not use it.

I did merge another pull request from another user so please rebase and we should be all set once this is addressed.  As always thanks for helping out :)

sandrapascual:
Hi @hazendaz, thanks for your advices. You were right about the concatenatios of the string. I've changed it. I've tried to push my commmit. You know about my "problems" with rebasing... please let me know if I have to change anything.

hazendaz:
@sandrapascual Looks like you still need to reformat using spaces instead of tabs.  In each file an easy way to do this without changing your eclipse setup is to run a regex search in the file and replace '\t' with '    '.

Rebasing is pretty each so lets see if we can get you there.

1) Make sure you sync'd with remote which you probably are.

git pull --rebase upstream master

2) Squash your commits into one.  When running this just follow directions to squash your commits.  This assumes the 4 you currently show is the same after the first step.  If that changes just adjust the numeric value to match what you want to rebase.

git rebase -i HEAD~4

3) Now you will have diverged from your fork.  So you need to force push.

git push origin +master

At this point, we should see this pull request still, one commit from you, and it fully in sync.  See if that helps get you there in how to do that.  Also please remember to switch out tabs for spaces :)

sandrapascual:
@hazendaz I've replaced tabs for spaces. Also, I've squashed the commits. Thanks for your recommendations. I think this is ok now.

hazendaz:
@sandrapascual Much closer...spacing is still off a bit.  We use 2 character spacing so there is quite a bit here that is off such that it seems more was changed than actually was.  Rebase is closer but still off a bit too.  

To get rebasing right.

git pull --rebase upstream master
git push origin +master

That will do it and you will have only one commit at that point.

To get spacing right, well, that is a tad more difficult than I think we would like right now.  I think we had formatted added once then it got removed again.  I'll try to add it back so we don't have to worry about these things as consistency is important.  Given your file set isn't too large, would you be so kind to get the spacing right?  If you have Eclipse and set that to two character spacing, a right click format will do the job then try to match back up the unchanged portions with original so we only see what you are bringing in.  After we get this, I'll work on getting formatting in place on a long term basis.

hazendaz:
@sandrapascual there are still formatting issues and non squashed commits.  However, I'm going to pick this by as I have re-introduced formatting and will have that hit before I grab this. I will push this in a separate pull request as soon as I can.  Thanks.

sandrapascual:
@hazendaz thanks! I'm not able to squash the commits. Sorry for the inconvenience.

hazendaz:
@sandrapascual Please see branch pr/734 for this pull request.  On there, you used a lot of string concatenization during logging.  That should bel avoided entirely.  You can write the logger similer to this 'logger.info("My nice logger {} takes two values {}', value1, value2)'.  This way there is no incurred cost of logging. If you can, check out this branch, make the changes, then submit a new pull request to master from that.  I'll squash the commits at that point and merge it.  (squash is now available within git!).

sandrapascual:
@hazendaz thanks for your advices. I've checked out the pr/734 branch and I've changed the logging lines. Now there is a new pull request so I'm closing this one. Thanks.

PR #746, opened 2016-07-16
Update deprecated jfreechart items

hazendaz:
@jander99 / @MALfunction84 please review this one.  It doesn't seem to cause any issues but I'm not very familar with what this piece of code actually does.

### DIFF
--- a/core/src/main/java/psiprobe/controllers/RenderChartController.java
+++ b/core/src/main/java/psiprobe/controllers/RenderChartController.java
@@ -162,9 +162,11 @@ public class RenderChartController extends AbstractController {
 
       final XYLine3DRenderer renderer = new XYLine3DRenderer();
       renderer.setDrawOutlines(true);
-      renderer.setLinesVisible(true);
-      renderer.setShapesVisible(true);
-      renderer.setStroke(new BasicStroke(2));
+      for (int i = 0; i < seriesMaxCount; i++) {
+          renderer.setSeriesLinesVisible(i, true);
+          renderer.setSeriesShapesVisible(i, true);
+          renderer.setSeriesStroke(i, new BasicStroke(2));
+      }
       renderer.setXOffset(1);
       renderer.setYOffset(1);
       chart.getXYPlot().setRenderer(renderer);


core/src/main/java/psiprobe/controllers/RenderChartController.java
```diff
@@ -162,9 +162,9 @@ protected ModelAndView handleRequestInternal(HttpServletRequest request,
 
       final XYLine3DRenderer renderer = new XYLine3DRenderer();
       renderer.setDrawOutlines(true);
-      renderer.setLinesVisible(true);
-      renderer.setShapesVisible(true);
-      renderer.setStroke(new BasicStroke(2));
+      renderer.setSeriesLinesVisible(0, true);
+      renderer.setSeriesShapesVisible(0, true);
+      renderer.setSeriesStroke(0, new BasicStroke(2));
```
MALfunction84: Just looking at the API, it kinda looks like these should be done in a for-loop now:

```java
for (int i = 0; i < seriesMaxCount; i++) {
  renderer.setSeriesLinesVisible(i, true);
  renderer.setSeriesShapesVisible(i, true);
  renderer.setSeriesStroke(i, new BasicStroke(2));
}
```

I'd also suggest testing it locally just to be sure it still works.

hazendaz: Thanks for the help.  Ill give the for loop a try.

Get Outlook for Android<https://aka.ms/ghei36>



On Sun, Jul 17, 2016 at 2:28 PM -0400, "Mark Lewis" <notifications@github.com<mailto:notifications@github.com>> wrote:


In core/src/main/java/psiprobe/controllers/RenderChartController.java<https://github.com/psi-probe/psi-probe/pull/746#discussion_r71086971>:

> @@ -162,9 +162,9 @@ protected ModelAndView handleRequestInternal(HttpServletRequest request,
>
>        final XYLine3DRenderer renderer = new XYLine3DRenderer();
>        renderer.setDrawOutlines(true);
> -      renderer.setLinesVisible(true);
> -      renderer.setShapesVisible(true);
> -      renderer.setStroke(new BasicStroke(2));
> +      renderer.setSeriesLinesVisible(0, true);
> +      renderer.setSeriesShapesVisible(0, true);
> +      renderer.setSeriesStroke(0, new BasicStroke(2));


Just looking at the API, it kinda looks like these should be done in a for-loop now:

for (int i = 0; i < seriesMaxCount; i++) {
  renderer.setSeriesLinesVisible(i, true);
  renderer.setSeriesShapesVisible(i, true);
  renderer.setSeriesStroke(i, new BasicStroke(2));
}

I'd also suggest testing it locally just to be sure it still works.

-

hazendaz:
I haven't proven this is working yet.  Just switched to eclipse neon and having debugging issues.  Will get back to this on another night but this does appear the right solution.

hazendaz:
After fixing spring/prototype related issue, I cannot see that this causes any issues.  Even the prior change I had in there I actually had been using in production setting for a while without any concerns.  I'm still having direct debugging issues but going to let this otherwise go.  All the charts look good otherwise.  The real key here was that as pointed out, it needed to loop through.

PR #756, opened 2016-07-21
[pom] Do not fail build on missing .git directory

hazendaz:


hazendaz:
Fixes issue raised on user group by allowing users to copy zip and use that to build rather than git clone.

PR #760, opened 2016-07-24
[jsp] Cleanup some jsp issues

hazendaz:


hazendaz:
Previously missed changes to get rid of warnings in jsp code.  All effectively are no difference changes.

PR #770, opened 2016-07-31
[git] Cleanup usage of git-commit-id-plugin

hazendaz:


hazendaz:
Older versions of this plugin didn't handle the git location very well.  That has long since been resolved so we no longer need to provide direction as to where the .git directory is.  Additonally since this is a well formed mulimodule project in git, we don't need this plugin running multiple times during a build.

PR #810, opened 2016-10-10
[clean] Add functional interface annotation where noted by sonar

hazendaz:


hazendaz:
closing... this is java 8 available only so premature in adding.

PR #825, opened 2016-11-19
[utils] Fix for singleThreadModel as it might be Boolean or boolean

hazendaz:
depending on tomcat version the return type was changed.  Technically speaking this is legacy servlet 2.4 code as it was for single thread check and there is no replacement as expected.  Tomcat internally changed how that method worked though as it was causing creation of instances if not already loaded for a class.  So this was fixed to return null when a class requested is not loaded.  So internally we are using reflection now and Null will simply be treated as false.  While that could be wrong based on context not being loaded, it's not really worth the effort given legacy here to make it smarter.


tachtler:
Hi Jeremy,

ich will try that as soon as possible and will give you feedback!

Thank you for your very fast analysis and patch for that problem - amazing support!
Klaus.

PR #879, opened 2016-12-14
Replace obsolete Java IP (InetAddress) Locator

hazendaz:
I initially wanted to entirely replace inetaddresslocator as it is from 2005 with database that old throwing accuracy into serious question.  While updates were made past (2.23), none are available in maven central, sourceforge library is still in CVS, and all updates noted as occurring up through 2007 are not even part of the codebase.  The download is otherwise manually available.

This left me with two alternatives.  The first was to try to resurrect the original library but I'd lose all history as it's not easy to get especially with no cvs experience and no easy import provided by github.  The other alternative was to look into other third party libraries.  This brought me to geo lite 2.

Our first use case was in Tomcat Application Utils.  I understand this first use case is under a valve so it's truely users using other applications other than simply logged into psi probe.

After looking at geo lite 2, I found it difficult to determine accurately the language which even earlier library was probably not exact at.  During this discovered that best place to get value is the Accept-Language header.  If missing tomcat, defaults it to system setting.  While it's possible user could change this, it seemed very viable solution given age of original and purpose to simply show a native flag.  Added bonus here it shows the flags for localhost as well.  So accepted this possibility as an improvement without additional libraries. 

Our last use case though proved to be a bit more difficult.  The code gets remote address out of jmx for use on the connectors page.  There is no way that I could see to introduce some library less way to do as I did for the tomcat valve solution.  Therefore, I added the support for geo lite 2.  Again as with the tomcat valve solution, I have this now showing the flag from system for localhost.  I additionally removed try/catch block for something not thrown as well as flag that was never used.

Please review and let me know if this looks good.


hazendaz:
One thing to note.  While geo lite didn't expliciently state it, jackson and httpclient were optional libraries.  One for JSON of course which we didn't need and the other for connecting to their webservice client instead of using downloaded database.  Obviously in our case downloaded data base is good enough with less hassle involved.

hazendaz:
Rebased and reworked.  Jackson turned out to be required.  Plus the city database is way too much for use case here so replaced simply with much smaller country database.

hazendaz:
@MALfunction84 @jander99 

I released a 3.0.0.RC1.  I think I'm going to merge this now that a release candidate is out of the way.  This adds to size of project but otherwise does what I think is right.  More concerned about size of addition than change really but would still like some feedback in case I didn't understand this properly or there is an easier way.  I'll merge likely later tonight.

And you might have guessed...I'm on an open source vacation so I've been working non stop on this.

MALfunction84:
How much bigger are we talking?

hazendaz:
Initially with city database it was double our current size.  After switching to country database it was negotiable but requires jackson which adds about 3 mb.  We are under 20 mb still and most likely if we get rest going we will be using jackson anyways.  Now if there is a way to capture info we need on jmx connectors we could drop geo.  I didnt see any other way there.  Possible we can figure it out though.

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

PR #971, opened 2017-01-03
[context] For parallel deployment of ROOT - Fixes #545

### LINKED ISSUE #275: Parallel Deployment on Tomcat 7 fails
```
Which existing feature would you like to see enhanced?
Applications Tab

Please describe the enhancement in detail.
There is a parallel deployment feture on Tomcat 7
http://tomcat.apache.org/tomcat-7.0-doc/config/context.html#Parallel_deployment
If one has ROOT.war application, and copy ROOT##001.war file with would be a 
newer version of the app, Tomcat would continue to server old sessions on the 
existing ROOT.war, but new sessions would go to the newer App. This way we 
could have Zero Downtime during deployment.
But if I deploy via Psi-Probe, I'm getting an error (even if application 
deployment works if I just copy ROOT##001.war to web apps folder). And if I 
copy the new app manually to web apps folder, Psi-Probe and everything works, 
but when I try to see Application Summary page of the new Application I'm 
getting error message instead,

How will this enhancement make Probe better?
It would support very important feature of Tomcat 7.

Thank you.
```

Original issue reported on code.google.com by `eug...@strokin.info` on 6 Mar 2012 at 11:42

### LINKED ISSUE #545: Parallel Deployment on Tomcat 7 still fails
When we use it to do Parallel Deployment on Tomcat 7,
something like ROOT##001.war,
it still fails.

hazendaz:
Fix context in safe context location

ROOT##1 style is treated differently with tomcat.  Our code was adding /ROOT##1 throughout.  While it lands on disk that way, tomcat doesn't see it that way.  It's really ##1.  So while it appeared issue was fixed for non ROOT usage, it was always broken for root usage.  Now it works for ROOT usage as well as normal usage.  ROOT entries will simply show as ##1 rather than ROOT##1 as on disk.

Entirely fixed best I can tell.  Please review and let me know if there are any concerns about where I fixed this.  Seems like all is ok but might want a little more testing tomorrow just to confirm no down side.

### DIFF
--- a/core/src/main/java/psiprobe/AbstractTomcatContainer.java
+++ b/core/src/main/java/psiprobe/AbstractTomcatContainer.java
@@ -266,6 +266,14 @@ public abstract class AbstractTomcatContainer implements TomcatContainer {
     if ("/".equals(result) || "/ROOT".equals(result)) {
       result = "";
     }
+    // For ROOT Parallel Deployment, remove "/ROOT"
+    if (result.startsWith("/ROOT##")) {
+      result = result.substring(5);
+    }
+    // For ROOT Parallel Usage, remove "/"
+    if (result.startsWith("/##")) {
+      result = result.substring(1);
+    }
     return result;
   }
 


hazendaz:
Sorry comments were incorrect.  Just fixed them and force pushed this back up ;)

Issue #889, opened 2016-12-15
Raise code coverage to 30%


hazendaz:

hazendaz:
closing as we are at 34% now.

PR #973, opened 2017-01-06
Switch spring security to annotation based - Fixes #811

### LINKED ISSUE #811: Spring annotations
Migrate all spring based xml to annotations.

This is not a straight port.  80% of the work I've completed but ran into some considerations.

1) Some tomcat code in the core needs to move.
2) We need to use abstract classes in many cases to clearly separate objects.
3) First pass results in many broken pieces like getting to logs due to pathing setup.

I've been on vacation for a few weeks.  I need to regroup on my current progress.  I expect as soon as I finish that I will be cutting a special branch for this work as it's huge.  At that point, others probably can jump in and give a helping hand to get this over the hump.  I think this effort will clearly open the doors for spring booting this as well.

hazendaz:
finishes work on #811.  All spring config now is java config based instead of xml.


hazendaz:
Works but big drop in test coverage.  Will work to bring that up on #889 as none of the annotation based code is getting covered.  I believe prior to final conversion here the bad unit test for certs was loading all the objects and showing coverage if I recall correctly.  That would make sense given the large drop.

PR #988, opened 2017-01-13
Add neutral flag

tachtler:
Add neutral flag as unknown.png for private ip address ranges on remote ip addresses or when something could not be determined.


hazendaz:
We might want to find an image that is a little more pleasing on the eyes.  Or possibly something that indicates private network.  Essentially the issue at hand right now is that on failure I made it use the system setup on so if it's USA flag then that is what is seen.  Probably not the best idea but me being in the states made sense to me.  Anyway, not forgetting about this but have been supper busy so if you have some additional time to maybe find something a little nicer to look at and potentially have the code change as well that would be great!

tachtler:
Hi Jeremy,

you can close that, I will search for a better frree picture to use.

Klaus.

PR #993, opened 2017-02-17
Faster Tomcat startup with JarScanner context.xml element.

jhorowitz-firedrum:


hazendaz:
@jhorowitz-firedrum Thanks for this change! Keep them coming!  I'm going to squash and add this in order to get rid of the merge commit.

PR #996, opened 2017-02-19
Fixes #898

diogosantana:
Fixes #898

It does not include PEM file configurations.


hazendaz:
@diogosantana Nice work!  I had taken a stab at this and clearly I was way off.  I'm merging now and will run some tests.

PR #998, opened 2017-02-19
Fixes #997

### LINKED ISSUE #997: Revert fbcd5714
I use Eclipse IDE Server's view to develop psi-probe, but with the default log configuration that Eclipse does to Tomcat together with Probe Logback configuration I can't see any log in console or a log file. That's why I created the logback-test.xml file back there.

diogosantana:
Fixes #997

### DIFF
--- a/web/src/test/resources/logback-test.xml
+++ b/web/src/test/resources/logback-test.xml
@@ -21,20 +21,15 @@
     </encoder>
   </appender>
 
+  <timestamp key="timestamp" datePattern="yyyyMMdd-HHmmss"/>
+
   <appender name="PSI-PROBE" class="ch.qos.logback.core.rolling.RollingFileAppender">
-    <File>${probe.log.path}/probe.log</File>
+    <File>target/temp/logs/probe-${timestamp}.log</File>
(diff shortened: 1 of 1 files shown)


hazendaz:
@diogosantana Makes sense.  I develop it a bit different using git bash and various tomcat versions outside of eclipse.  Can you confirm this file doesn't end up in the resulting WAR after build?  I'm thinking it might and will need an exclusion on the war plugin.  I think there might have been an exclusion before but haven't looked.  I'll merge once you confirm that.  Thanks.

hazendaz:
@diogosantana I"m going to go ahead and merge this since I'll be testing.  Since this mainly affects you if you could still confirm our ending file doesn't end up with a test file.  I haven't used eclipse in so long to test that I wouldn't know where to begin.  Even when I use the servers there, i have full control of installed tomcat so it uses the normal logging.  Thanks.

hazendaz:
@diogosantana No need to verify anything.  Didn't realize it was in the test package.  All good.  It's back in the code base so you should be good to go going forwards.  Thanks.

PR #1006, opened 2017-02-19
[log4j2] Add change log level and update some comments

hazendaz:


hazendaz:
@jhorowitz-firedrum Can you take a look at this and see if it makes sense to have a change log level for log4j2?  I had made an original attempt for log4j2 but it wasn't finished and I didn't quite know how to test it.  Anyway besides you used classes I didn't quite know about which I fealt was much better than my attempt this was the only thing I had left over that I wasn't quite to sure about dropping.  Let me know if this makes sense and if it does I'll merge this in.  I probably need your help to test and confirm.

jhorowitz-firedrum:
The comment changes you made to the LogResolverBean missed a couple of "log4 j" -> "log4j" replacements.

Just tested with the native log4j2 xml automatic reconfiguration, and using your code the change seems to persist as long as the configuration hasn't been updated. However, an update to the log4j2 configuration source will overwrite the change.

I think that behavior is preferable though... what do you think?  I'd say, go ahead and merge. :smile:

hazendaz:
Thanks @jhorowitz-firedrum.  Yes the change within psi probe is expected to be temporary in nature.  Thank you for testing.  I caught the extra javadocs.  I'll merge shortly once travis-ci reports back success.

PR #1007, opened 2017-02-20
Fix live update of log4j2 log levels.

jhorowitz-firedrum:
Run code formatter for log4j2 classes.

Signed-off-by: Jonathan Horowitz <jhorowitz@firedrum.com>


jhorowitz-firedrum:
Sorry @hazendaz! My earlier test of your controller change was only half accurate. It was indeed updating the level of the in-memory configuration, but apparently there was another call required to push that change out to the actual configured loggers.

This PR adds that "updateLoggers" call on the context after setting the level.

hazendaz:
Np. Thanks for the PR.

PR #1009, opened 2017-02-20
[cert] Certificates tab works again in Tomcat 7/8. Fixes #1000

### LINKED ISSUE #1000: Tomcat 7/8 no longer show cert information
Patches to support org/apache/tomcat/util/net/SSLHostConfig cause failure with tomcat 7/8.  We need to allow for backwards compatibility.

diogosantana:
Fixes #1000.

I don't know why Eclipse changed import orders...

### DIFF
--- a/core/src/main/java/psiprobe/controllers/certificates/ListCertificatesController.java
+++ b/core/src/main/java/psiprobe/controllers/certificates/ListCertificatesController.java
@@ -10,24 +10,6 @@
  */
 package psiprobe.controllers.certificates;
 
-import org.apache.catalina.connector.Connector;
-import org.apache.commons.beanutils.MethodUtils;
-import org.apache.coyote.ProtocolHandler;
-import org.apache.coyote.http11.AbstractHttp11JsseProtocol;
-
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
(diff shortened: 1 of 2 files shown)


hazendaz:
alt-shift-o to import will cause eclipse to re-order things.

hazendaz:
@diogosantana This does not fix the issue.  Did you test this solution?  I'm using tomcat 8.0.39, and it gets an exception internally and doesn't set the keycerts.  My thought to fixing was to just populate the new objects you created with the legacy information.  At the moment, I still get a blank result.

hazendaz:
specifically -> keyStoreCerts of certificateInfos is null.  Therefore it won't be able to be displayed on the webpage as it expects the new layout.  I'm not sure an extension to connectorInfo makes sense here considering internal certificate info is still setup and that is what the page wants.  We are just faking tomcat out here in this case so we need to mirror into into it as if the object really exists.

hazendaz:
I got it...needed null checks from original code and could use the same object not changing object type which isn't otherwise used later on.  PR coming shortly.

diogosantana:
Ok, thanks.

I know about organize import shortcut, but eclipse change to that order... I don't know if Neon changed the default order or what... anyway, it really doesn't matter.

PR #1075, opened 2017-06-14
Update dependencies

hazendaz:


hazendaz:
Reverting jmockit for time being.

PR #1083, opened 2017-07-03
[pom] Add error prone library usage with compiler

hazendaz:


hazendaz:
error prone requires java 8, so closing this as not necessary as it catches no errors currently.  I'll keep this local.

Issue #1097, opened 2017-08-14
tomee data souce not listed in Data Sources tab


dmcletchie:
Hi,

I hope that this isn't a known issue or a config problem.

I'm using tomee, and I'm defining a datasource.  If I define it in a context.xml file, then PSI probe works as expected.  I see the data source in the global data source tab.  if I define it tomee.xml, then the datasource is not listed.  The resource does appear in applications -> resources.  I can click on it & test the connection & it works. 

I'd like to see the data source under the global datasources tab, so that I can monitor the use of my connection pool.  I've debugged & fixed the problem locally

It looks like the TomcatJdbcPoolDatasourceAccessor should work, except that the resource class is not a "org.apache.tomcat.jdbc.pool.DataSource", it is a "org.apache.tomee.jdbc.TomEEDataSourceCreator$TomEEDataSource".  the TomEEDataSource extends the DataSource.   TomcatJdbcPoolDatasourceAccessor isn't working since it's checking the class name, and it does not match for a subclass.   So, I could either:
* modify TomcatJdbcPoolDatasourceAccessor to include the second class name, 
* remove the class name comparison leave the instanceof as the only check
* create a new DatasourceAccessor just like TomcatJdbcPoolDatasourceAccessor  but it a different string constant in canMap().  

(If nobody suggests a preference, I guess I'll do the first one.)

I'm new to github, so I'm not sure what protocol is or how to go about getting this fix in place.  Any pointers would be great!  Thanks!

hazendaz:
New class would be best.  Your option 3.  Feel free to make fix in your fork then send a pull request.  I'll merge it fairly quickly.  Thanks for finding the issue.

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

dougwm:
I have the same problem if I configure a dbcp2 datasource inside of tomee.xml.  So, if you're OK with the way that this one works, I'll create another issue & another pull request.

hazendaz:
Hmmm...now your option 1 is sounding better.  If you want go ahead and send a PR covering both for me to look at.  Option 1.  Then we can go from there.

Get Outlook for Android<https://aka.ms/ghei36>



________________________________

hazendaz:
@dougwm I'm good with what you did.  Go ahead and raise a pull request to provide support for dbcp2 usage.

hazendaz:
long addressed, closing.

PR #1098, opened 2017-08-14
for issue 1097 

dougwm:
This is my first experience with github, so I might not be doing this correctly.  I amended my commit since I forgot to add the second file.  
I can easily redo this.  Let me know if I should do something differently.

Thanks!

PR #1099, opened 2017-08-15
 add tomee datasourceAccessor issue #1097  (#1098)

dougwm:
fixes issue #1097 
Thanks for your help.  I tried to follow what you said, but I'm pretty sure that I got something wrong.

I added the main repository as an upstream & did a hard reset & pulled from it.  The I added a few more changes for the TomEE dbcp2.  I ammended the commit.  I did something with a rebase to squash the 2 commits into 1, but I think that's where I got lost.  Sorry to be such a rookie ... i think that I squashed yesterday's change & today's change into one commit, pushed it, but when I create the pull request it says "Can’t automatically merge".   probably because I have a different change ID here since I squased them into 1. (so my repo doesn't have the unsquashed one?)

What's easiest for you?  Should I create a new ticket for this change set?  Sorry that I don't know what I'm doing with github / git.  I really appreciate the help.

### DIFF
--- a/core/src/main/java/psiprobe/ProbeConfig.java
+++ b/core/src/main/java/psiprobe/ProbeConfig.java
@@ -158,6 +158,8 @@ public class ProbeConfig extends WebMvcConfigurerAdapter {
     list.add("psiprobe.beans.accessors.OracleUcpDatasourceAccessor");
     list.add("psiprobe.beans.accessors.OpenEjbManagedDatasourceAccessor");
     list.add("psiprobe.beans.accessors.ViburCpDatasourceAccessor");
+    list.add("psiprobe.beans.accessors.TomEEJdbcPoolDatasourceAccessor");
+    list.add("psiprobe.beans.accessors.OpenEjbBasicDatasourceAccessor");
     return list;
   }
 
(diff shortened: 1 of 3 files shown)


hazendaz:
I think the class name you should use should be 'OpenEjbBasicDatasourceAccessor' since it looks like it is actually out of openEjb rather than tomee directly.

To fix your local, all you need to do is the following

```git pull --rebase upstream master```

That should flush out the changes that were already merged.

Then just do this...

```git push origin +master```

If you do change the class to better indicate the source, you might as well just soft reset the head after you have done the above.

So for example, you have done the above and now have one commit with just the changes yet to be merged, then you apply one more commit changing the name, you would do this.

```git reset --soft HEAD~2```

If you had not commited the file change name yet, then it would have been a 1 instead of 2.  Then again, when pushing back up, use the force push (ie git push origin +master).  That should get you where you need to be.  You won't even have to do a thing on this pull request as fixing it that way will in turn fix the pull request so it shows it can be merged.

dougwm:
I took a long & winding path, but I think that I got where I needed to get.  Thanks for your help with this.  If anything is still out of place, please let me know.

hazendaz:
@dougwm Thanks.  I'm still going to squash as you have a merge conflict on there.  I know learning in progress ;)  So I'm good with this.  Thanks.

PR #1104, opened 2017-09-05
Update dependencies

hazendaz:


hazendaz:
This passed.  Java 7 variation fails on trusty for travis so ignoring until they fix it.

PR #1119, opened 2017-11-28
Feature/java util logging filehandler

### LINKED ISSUE #363: Standard JDK logging handlers not seen
```
1. in Tomcat's logging.properties configure some handler with 
java.util.logging.FileHandler
2. go in browser to PsiProbe in the "logs" tab

What is the expected result?
a line for for the FileHandler should be shown, allowing access to the 
associated file contents, just as it would be with org.apache.juli.FileHandler.

What happens instead?
Nothing appears if it's the only handler; Maybe some console handlers if some 
are configured also.

2.3.2

any browser, Tomcat 7.0.27, JRE 1.6.0.33, RHEL 6


(Attach logs or stack traces as files instead of pasting the contents
here.)

Looking at the sources I found in Jdk14LoggerAccessor that it will only 
recognize java.util.logging.ConsoleHandler and org.apache.juli.FileHandler;
It should be more generic and accept instances of (sub)classes within
- java.util.logging.FileHandler
- java.util.logging.ConsoleHandler
- java.util.logging.MemoryHandler
at least.

We've got some Java extensions which define new handler classes deriving from 
the standard classes, so subclasses should definitely be accepted.

Maybe there's a way to be even more generic accepting Handler directly, but I 
guess not in a useful way.
```

Original issue reported on code.google.com by `fabi1cou...@gmail.com` on 23 Jul 2013 at 12:50

thberger:
Fixes #363 for java.util.logging.FileHandler

Note: Did not change Jdk14HandlerAccessor's interface, so only the latest log file is taken into account, even if multiple files are rolled over by the FileHandler.

### DIFF
--- a/core/src/main/java/psiprobe/tools/logging/jdk/Jdk14FileHandlerAccessor.java
+++ b/core/src/main/java/psiprobe/tools/logging/jdk/Jdk14FileHandlerAccessor.java
@@ -1,0 +1,30 @@
+/**
+ * Licensed under the GPL License. You may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   https://www.gnu.org/licenses/old-licenses/gpl-2.0.html
+ *
+ * THIS PACKAGE IS PROVIDED "AS IS" AND WITHOUT ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING,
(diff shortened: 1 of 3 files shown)


hazendaz:
@thberger 

Great!  Looks good.  Thank you so much!   Not sure what the deal with travis is but I'm merging it.

hazendaz:
ok - just confirming tavis is having issues.  Merging now 👍

PR #1125, opened 2018-01-01
All things java 9 now building

hazendaz:
jspc -> Not even sure why original was even used as it was not compiling so unsure what it's purpose was.  I think it should have compiled to confirm code was good then dumped the compilation.  Anyway, this will now compile the jsp into a directory in target 'jspc' and then will simply not pick it up for distribution.

Remainder brought up to java 9 build support.  For build related items like the jetty piece I'm going to have to rework a bit to allow further upgrades but think we might drop java 7 soon as it's becoming too complicated to keep supporting the past and with nearly a million downloads a release building on java 9 is a good point to just start diverging from that past.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ language: java
 sudo: false
 
 jdk:
+  - oraclejdk9
   - oraclejdk8
   - openjdk8
   - openjdk7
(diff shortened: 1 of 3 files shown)


hazendaz:
travis has been horrible all day.  Java 9 is confirmed on my machine after much work.  I'm merging this as travis jobs are not even starting now and getting quote errors.  Maven was also out today which hasn't helped things :(

PR #1173, opened 2018-05-17
Update dependencies

hazendaz:


hazendaz:
This should be fine in a couple of hours.  Merging.

Issue #1140, opened 2018-02-26
Tomcat 8.5 ResourceLink not shown


claudioweiler:
Global resources are correctly shown on Data Sources, but links do not.

Resources info inside application says that all resources are misconfigured.

But application that use those links works without problems.

claudioweiler:
It appears that Tomcat created a access validation to ResourceLinks.
I modified Tomcat85ContainerAdapter.addContextResourceLink to register access, it need some improvements in overall AbstractTomcatContainer structure.

```
  public void addContextResourceLink(Context context, List<ApplicationResource> resourceList,
      boolean contextBound) {

    NamingResourcesImpl namingResources = context.getNamingResources();
    for (ContextResourceLink link : namingResources.findResourceLinks()) {
      ApplicationResource resource = new ApplicationResource();

      logger.debug("reading resourceLink: {}", link.getName());
      resource.setApplicationName(context.getName());
      resource.setName(link.getName());
      resource.setType(link.getType());
      resource.setLinkTo(link.getGlobal());

      ResourceLinkFactory.registerGlobalResourceAccess(getGlobalNamingContext(), link.getName(), link.getGlobal());

      resourceList.add(resource);
    }
  }
```

hazendaz:
@claudioweiler Does this work?  If so, can you add to a PR and also patch tomcat9.

hazendaz:
@claudioweiler Lets see what you can create and go from there.  I'm not sure which piece you are waning in the resource resolver.  The fields / methods are plublic and/or static where needed already.  I may be missing exactly what you want changed for this but I'm more than happy to work with you on this.  If we need any of those changes that should be fine to accomplish this.

claudioweiler:
Ok @hazendaz, I will make it my way, but I can't right now.

I've made a hack here just for make it work, no good to create a PR from that.

Tomcat 8.0 will need it too?

hazendaz:
Fixed via #1176

PR #1176, opened 2018-06-12
issue_1140

claudioweiler:
solves #1140

### DIFF
--- a/core/src/main/java/psiprobe/AbstractTomcatContainer.java
+++ b/core/src/main/java/psiprobe/AbstractTomcatContainer.java
@@ -24,12 +24,14 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+
 import javax.management.MBeanServer;
 import javax.management.MalformedObjectNameException;
 import javax.management.ObjectName;
 import javax.naming.NamingException;
 import javax.servlet.ServletConfig;
 import javax.servlet.ServletContext;
+
 import org.apache.catalina.Container;
 import org.apache.catalina.Context;
 import org.apache.catalina.Engine;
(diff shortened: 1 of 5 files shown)


core/src/main/java/psiprobe/beans/ResourceResolverBean.java
```diff
@@ -308,10 +311,10 @@ private String getStringAttribute(MBeanServer server, ObjectName objectName,
    *
    * @return the global JNDI context
    */
-  protected javax.naming.Context getGlobalNamingContext() {
+  public static javax.naming.Context getGlobalNamingContext() {
 
     javax.naming.Context globalContext = null;
-    MBeanServer mbeanServer = getMBeanServer();
+    MBeanServer mbeanServer = ManagementFactory.getPlatformMBeanServer();
```
hazendaz: @claudioweiler Smart...no reasont to call getter in same class to perform the same action.

core/src/main/java/psiprobe/AbstractTomcatContainer.java
```diff
@@ -467,6 +473,14 @@ public void unbindFromContext(Context context) throws NamingException {
     changeContextBinding(context, false);
   }
 
+  /**
```
hazendaz: @claudioweiler Great!  Simple enough.  Thanks for the add.  I'm merging now.

core/src/main/java/psiprobe/beans/ResourceResolverBean.java
claudioweiler: @hazendaz 
To use the getter it need to be changed to static too, has impact on interface and others implamentations.

PR #1208, opened 2018-10-13
Bring build up-to-date supporting jdk11 (mostly) with error prone rework

hazendaz:


hazendaz:
jdk 11 support still has issues with awt error leak check.  I'm going to see about getting classloader leak prevention working instead and simply remove that.

Issue #1275, opened 2019-03-20
wrapper Request method 'POST' not supported


mrchuseau:
Hi, I am using tomee plume 7.1.0 with java service wrapper

Psi-probe Version 3.2.1-SNAPSHOT

In System tab Wrapper control menu
If i click any of the icons
Request thread dump
stop JVM
restart JVM

I get the HTTP Status 405 – Method Not Allowed

in probe.log
WARN  {http-nio-8080-exec-16} [o.s.w.s.m.s.DefaultHandlerExceptionResolver] Resolved [org.springframework.web.HttpRequestMethodNotSupportedException: Request method 'POST' not supported]

![image](https://user-images.githubusercontent.com/4137782/54707843-c94fd600-4b20-11e9-934a-c0932ce995c6.png)

hazendaz:
@mrchuseau Is it possible for you to provide some basic instructions and/or sample to setup your use case?  Or if feasible provide a bit more of the stack trace that should be in your logs from psi probe so I can figure out where this issue exists?

mrchuseau:
OS: Centos 7.6.1810
java: oracle jdk 1.8_201
It's a standard tomee plume installation 7.1 unzipped from the homepage
java service wrapper https://wrapper.tanukisoftware.com/
integration method 2 https://wrapper.tanukisoftware.com/doc/english/integrate-start-stop-nix.html
modify tomcat_user.xml
`<user username="tomcat" password="masked" roles="manager-gui,admin-gui,monitoring,tomee-admin"/>`
`bin/tomee installstart`
copy probe.war to webapps

I have attached the log

[probe.log](https://github.com/psi-probe/psi-probe/files/2991817/probe.log)

hazendaz:
I'll try to look at this soon.  Its probably an easy fix if you have time to look at it.  At the moment I'm not sure exactly where this is tripped up but whe moving to newer spring versions I had to do a few funny things like add 'PostParameterizableViewController' to restore POST usage.  It's possible this might be an edge case in that area that you might be able to track down quicker and raise a pull request to patch.

hazendaz:
Fixed per #1279.

PR #1279, opened 2019-03-26
Fixes issue #1275. 

dousdebes:
Classes in wrapper package now extends from PostParametrizableViewController to support POST method.


hazendaz:
@dousdebes Thanks!  I'll try to get a release pushed in next week or so.

PR #1534, opened 2020-02-26
Show some cluster information when it has been configured in Tomcat

gorbeia:
I have tested this in Tomcat 9 and shows some information.

### DIFF
--- a/core/src/main/java/psiprobe/beans/ClusterWrapperBean.java
+++ b/core/src/main/java/psiprobe/beans/ClusterWrapperBean.java
@@ -12,9 +12,11 @@ package psiprobe.beans;
 
 import java.lang.management.ManagementFactory;
 import java.util.Set;
+
 import javax.management.MBeanServer;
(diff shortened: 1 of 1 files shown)


hazendaz:
What about tomcat 7 and 8.5?  Also, if changed,  I would prefer commented out code to be deleted.  I don't use clusters myself and know support was weak here.  Do you have any information that points to this in tomcat to confirm validity?  I don't know otherwise...

gorbeia:
I am checking a Tomcat 8.5 instance an MBean names are different. Some version checking will be needed to support several Tomcat versions.

What is the right way to do version checking in psi-probe. Is it advisable to retrieve the version somehow in ClusterWraperBean and add a few if/elses to use the right bean names? How can we check the Tomcat version?

hazendaz:
closing as never followed up on and some code was commented out.  probably worth looking at but not in this current state which isn't being reviewd any longer.

PR #1567, opened 2020-04-03
[travis] Use jdk 14 intead of 12

hazendaz:


hazendaz:
Will address removing error prone later since google seems incapable of actually mering fixes that properly support 12/13/14/15/etc.

PR #1613, opened 2020-05-24
Fix the connector chart cannot slide to the maximum value

Lucien11:
Fix the connector chart cannot slide to the maximum value, and fix the bug that the slider chart cannot be changed

PR #1614, opened 2020-05-24
[pom] Correct spacing with javax.mail exclusion

hazendaz:


hazendaz:
Noted on PR #1613 .  Applying separately as unrelated to that change.

PR #1616, opened 2020-05-25
Fix the bug that the connector chart cannot slide to the maximum value, and fix the bug that the chart size cannot be changed by clicking the slider

Lucien11:
Fix the bug that the connector chart cannot slide to the maximum value, and fix the bug that the chart size cannot be changed by clicking the slider
image
image


hazendaz:
@Lucien11 This looks good to me.  I had just released a new release less than a week ago.  If its not a rush for something you need now, can we wait around a month?  I don't mind either way, I'll let you decide.

Lucien11:
> @Lucien11 This looks good to me. I had just released a new release less than a week ago. If its not a rush for something you need now, can we wait around a month? I don't mind either way, I'll let you decide.
ok, you can be merged into the next version,thank you

Issue #1332, opened 2019-06-04
Probe 3.2.0=>3.3.1 - Start application works only from main homepage, not from application dedicated page


aurelienpernoud:
Hello,

we started using Probe 3.2 and noticed that the "start" button when clicking on a stopped application and that uses another link that the "start" icon directly on home page doesn't work.

In other words you click on "manager" app for ex, Stop it using the top button "Stop", it works.

You try then to start it and it fails with 
"
There was an error

Application encountered an unexpected error. Please check the Tomcat logs. We will greatly appreciate it if you share the information with us. Sharing this information through our issue tracker or our discussion board will help us find and correct this problem as soon as possible!
"

Sample URL :
{host}/probe/app/start_summary.htm?webapp=%2fmanager

aurelienpernoud:
I noticed this error in the logs every time I click "start", we're using Tomcat 9.0.20 :

04-Jun-2019 15:30:32.064 SEVERE [http-nio-8080-exec-19] org.apache.catalina.core.StandardWrapperValve.invoke Servlet.service() for servlet [probe] in context with path [/prb] threw exception [Request processing failed; nested exception is java.lang.IllegalStateException: No primary or default constructor found for interface org.apache.catalina.Context] with root cause
	java.lang.NoSuchMethodException: org.apache.catalina.Context.<init>()
		at java.lang.Class.getConstructor0(Class.java:3082)
		at java.lang.Class.getDeclaredConstructor(Class.java:2178)
		at org.springframework.web.method.annotation.ModelAttributeMethodProcessor.createAttribute(ModelAttributeMethodProcessor.java:216)
		at org.springframework.web.servlet.mvc.method.annotation.ServletModelAttributeMethodProcessor.createAttribute(ServletModelAttributeMethodProcessor.java:84)
		at org.springframework.web.method.annotation.ModelAttributeMethodProcessor.resolveArgument(ModelAttributeMethodProcessor.java:139)
		at org.springframework.web.method.support.HandlerMethodArgumentResolverComposite.resolveArgument(HandlerMethodArgumentResolverComposite.java:126)
		at org.springframework.web.method.support.InvocableHandlerMethod.getMethodArgumentValues(InvocableHandlerMethod.java:166)
		at org.springframework.web.method.support.InvocableHandlerMethod.invokeForRequest(InvocableHandlerMethod.java:134)
[...]

PR #1649, opened 2020-07-07
Fix 1332

aurelienpernoud:
https://github.com/psi-probe/psi-probe/issues/1332


hazendaz:
@aurelienpernoud The method signature has been deleted here so it no longer compiles.

aurelienpernoud:
I don't know how it happened... my file had it locally, and now github dekstop asked me to commit again... just did it :-(

hazendaz:
@aurelienpernoud Thanks!  I did a check through code to see if anything similar.  This appeared to be the only one like this and clearly was incorrect.  Appreciate the fix and no worries on the git issue, you will get a hold of it :)

PR #2114, opened 2022-04-19
make build reproducible

hboutemy:
see https://github.com/jvm-repo-rebuild/reproducible-central/tree/master/content/com/github/psi-probe
too much details on the environment is adding some unnecessary noise


hazendaz:
@hboutemy Assume I should remove the extra build timestamps from the site plugin too.  The diff shows them still here.  I can merge this and follow up, can you confirm?

hboutemy:
site is not taken into account for reproducibility check, then you can keep it if you want.
But it's true that being consistent make sense

Issue #2290, opened 2022-11-01
ExceptionInInitializerError on Tomcat 9 and OpenJDK 17


sknoefel:
Environment:
OS: Debian GNU/Linux 11 (bullseye)
Tomcat: 9.0.43-2~deb11u3
Java: Open JDK 17.0.4+8-Debian-1deb11u1
Probe Version: 3.6.0 (also happens with 3.5.8)

Description:
We switched to Open JDK 17 (from 11).
When starting the Probe webapp first time it works well. When restarting the Tomcat server, the Probe webapp initialization failed with exception below and the webapp is down. When I manually delete tomcat/work/Catalina/localhost/probe I can start the webapp in tomcat manager and it works well again, until next Tomcat restart.

It looks like the loading of serialized statistics data on application initialization does not work with Open JDK 17.

`ERROR {WrapperStartStopAppMain} [o.s.web.context.ContextLoader] Context initialization failed
org.springframework.beans.factory.UnsatisfiedDependencyException: Error creating bean with name 'renderChartController': Unsatisfied dependency expressed through field 'statsCollection'; nested exception is org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'statsCollection' defined in psiprobe.ProbeConfig: Invocation of init method failed; nested exception is java.lang.ExceptionInInitializerError
        at org.springframework.beans.factory.annotation.AutowiredAnnotationBeanPostProcessor$AutowiredFieldElement.resolveFieldValue(AutowiredAnnotationBeanPostProcessor.java:659)
        at org.springframework.beans.factory.annotation.AutowiredAnnotationBeanPostProcessor$AutowiredFieldElement.inject(AutowiredAnnotationBeanPostProcessor.java:639)
        at org.springframework.beans.factory.annotation.InjectionMetadata.inject(InjectionMetadata.java:119)
        at org.springframework.beans.factory.annotation.AutowiredAnnotationBeanPostProcessor.postProcessProperties(AutowiredAnnotationBeanPostProcessor.java:399)
        at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.populateBean(AbstractAutowireCapableBeanFactory.java:1431)
        at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.doCreateBean(AbstractAutowireCapableBeanFactory.java:619)
[...]

PR #2339, opened 2022-11-25
[stats] Add catch on exceptionInInitializerError and provide user som…

hazendaz:
…e assistance


hazendaz:
Alternative fix for #2290 that helps the user understand they need add opens and how to do so since its obscure in java.  I'm unable to cause this issue myself but multiple users have reported it.  The original fix was too broad on the exception caught as throwable without enough rational as to why violate java conventions as many tools state that is far too broad and should not be caught directly.  One could say if we go broad, we do it everywhere which would be wrong.  Given I cannot even reproduce, going that broad doesn't make much sense here.  Also, in some future event going to replace xstream with usage of jackson to get same behavior given we have that already and it would work just as well which then probably means this goes away.  For now, this feels good enough.

PR #2535, opened 2023-05-20
[cleanup] Rework log4j2 to get rid of try catch around self exception…

hazendaz:
… throw


hazendaz:
missed part of the throw, reworking so no throws.

PR #2542, opened 2023-05-21
[logs] Add gc.log and localhost-access.log to stdout

hazendaz:
given no support for these, add them as typical naming


hazendaz:
Long asked for, easy fix.  gc logger and localhost-access both added to stdout.  These are most typical naming for both.

Issue #150, opened 2015-03-16
Refresh uptime via AJAX


GoogleCodeExporter:
```
Which existing feature would you like to see enhanced?
User Interface, display of uptime in page header.

Please describe the enhancement in detail.
The uptime is currently static in every page.  Add a simple AJAX call which 
refreshes the uptime every minute.

How will this enhancement make Probe better?
Uptime information will always be up-to-date, never requiring a refresh.
```

Original issue reported on code.google.com by `MALfunct...@gmail.com` on 25 Sep 2010 at 2:48

hazendaz:
resolved by #2640

PR #2640, opened 2023-07-16
Add ajax onto uptime and fix minor issue passing 'null' that would default to 'null' already (sonar issue)

hazendaz:


hazendaz:
fixes https://github.com/psi-probe/psi-probe/issues/150

PR #2743, opened 2023-11-05
Sync javax to master

hazendaz:


hazendaz:
note drop in one library is to test dual renovate runs which seems working already but to be sure.

PR #2744, opened 2023-11-05
Allow test scope to be any version on transients / bump logback to 1.3.11 since renovate fails to do so on that one.

hazendaz:


hazendaz:
renovate issue is only due to the default usage, we would need to make it do patches separately.

PR #2857, opened 2023-11-19
[pom] Fix CVE issues with json smart

hazendaz:


hazendaz:
while most libs are controlled in parent, didn't bother here as we never implemented any rest code and I just haven't deleted this module yet.  So fixing it for time being.

PR #2875, opened 2023-11-24
Add mockito, add byte-buddy agent to argLine, and convert over first module from jmockit to mockito

hazendaz:


hazendaz:
This module was expected to remain same on coverage.  With the other set I'm switching out I expect some increase.

PR #2894, opened 2023-11-25
Adjustments to badges / links on readme and fix deprecated item on sonar

hazendaz:


hazendaz:
note we had 2.4k or so downloads on 3.7.3 release, I did not have 4.0.0 war pushed up, it is now so it is starting at zero.

PR #2914, opened 2023-12-01
Cleanup object handling to prevent external object manipulation of maps / lists

hazendaz:


hazendaz:
ok codeQL is correct, leaving for now as tracking those separately.  Did not need the checks here as object set at class level and nothing allows it to be null as indicated.

PR #3148, opened 2024-03-02
Issue 1961: Wait few seconds during upload to avoid empty context

ralf-devop:
As i described in issue 1961, context is empty if you don't wait.

### DIFF
--- a/psi-probe-core/src/main/java/psiprobe/controllers/deploy/UploadWarController.java
+++ b/psi-probe-core/src/main/java/psiprobe/controllers/deploy/UploadWarController.java
@@ -49,6 +49,7 @@ public class UploadWarController extends AbstractTomcatContainerController {
 
   /** The Constant logger. */
   private static final Logger logger = LoggerFactory.getLogger(UploadWarController.class);
+  private static final int MAXSECONDS_WAITFOR_CONTEXT = 10;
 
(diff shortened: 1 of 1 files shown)


hazendaz:
@ralf-devop Do you know if there is anything in tomcat we could better wait on then a delay like this?  If not sure, that's fine, I'll merge this in.  I'm just thinking there might be a better way to escape the process startup logic needed.  Anyway thank you, I can push a release today even on this.  I want to make sure our 'javax' usage is pretty sound before I have to sort of gut that same area so it works with jakarta as tomcat deleted the original code we are relying on.

ralf-devop:
In my opinion something like a callback is the better solution.

hazendaz:
coveralls is flaky on this repo so it does show 1 to 2 percent drops occasionally which has nothing to do with changes.  Want to hear back from @ralf-devop on if this is best we can do for now before I merge.  Will however as noted work on release right after so this is out in few hours.  This certainly won't hurt regardless.

hazendaz (approved):

ralf-devop:
I know it is just a quickfix but works for me so far.

hazendaz:
Thanks.  I should have this out in a few hours.  Working on some other stuff right now but will get it by 5pm or so at latest.

PR #3157, opened 2024-03-10
Issue 3154: Fixed size param position in requests

ralf-devop:
1. Setting size param in toggle tag at the end
2. Certain links (including ajax-calls) in appsummary.jsp have size parameter, now

### DIFF
--- a/psi-probe-core/src/main/java/psiprobe/jsp/ParamToggleTag.java
+++ b/psi-probe-core/src/main/java/psiprobe/jsp/ParamToggleTag.java
@@ -36,19 +36,29 @@ public class ParamToggleTag extends TagSupport {
     boolean getSize =
         ServletRequestUtils.getBooleanParameter(pageContext.getRequest(), param, false);
     StringBuilder query = new StringBuilder();
-    query.append(param).append('=').append(!getSize);
+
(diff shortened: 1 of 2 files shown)


hazendaz:
Will try to get to this in the coming weekend.

hazendaz:
@ralf-devop Sorry for delays, working on a few other items (some related here with getting some milestones off milestones on submodules I'm support - displaytag / sitemesh). Can you help me understand here though what might have caused this issue and possibly last point it was working?  I just want to be sure i understand the issue as on first glance it seems odd that attributes need to be in some order.  I'm ok with it otherwise just wanted a bit more understanding.  It may be about 2 weeks before I can get this out.  So this would at least help me out getting to that point :)

ralf-devop:
@hazendaz
With a little distance, I would say the real reason for this issue is the missing size parameter in some ajax-calls on appsummary.jsp.
We don't need any changes on toggle tag (ParamToggleTag.java). My first approach was comparing the the links. If you prefer uniform calls with leading webapp parameter,  then you could take my changes on ParamToggleTag.java.

hazendaz:
THanks @ralf-devop, that helped my understanding.  I've merged it.  I prefer how you did it as it seems more logical overall.

PR #3158, opened 2024-03-14
[pom] Move to jetty jspc 12.0.7 for ee8

hazendaz:


hazendaz:
jetty before 12 is no longer community supported.  version 12 was made to support ee8, ee9, and ee10 easily.  It will also support ee11.  For our needs, we only need the jspc part and it works without changes other than update.

hazendaz:
Requires to drop jdk 11 from build cycle.  Moving to do that.

hazendaz:
byte buddy claims to support jdk 23.  we have it locked to dependency management.  It shows in eclipse at right version in the tomcat 8.5 module where its failing.  Don't know what is not working there.  will come back to that issue.

PR #3239, opened 2024-04-19
Update tomcat and rework upload logic

hazendaz:
It seems tomcat backported changes to tomcat 9 that breaks same issue I faced with tomcat 10.1 usage.  I have hope pulling some of their code and confirming it still uploads might make jakarta a go.  Getting this merged then will retest.  If it works out, expect to release this again on javax then follow up a quick jakarta.  My hope anyways..

### DIFF
--- a/psi-probe-core/src/main/java/psiprobe/controllers/AbstractTomcatContainerController.java
+++ b/psi-probe-core/src/main/java/psiprobe/controllers/AbstractTomcatContainerController.java
@@ -10,8 +10,12 @@
  */
 package psiprobe.controllers;
 
-import javax.inject.Inject;
+import java.util.Locale;
 
+import javax.inject.Inject;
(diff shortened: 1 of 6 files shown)


psi-probe-core/src/main/java/psiprobe/controllers/AbstractTomcatContainerController.java
```diff
@@ -70,4 +80,24 @@ public String getViewName() {
   public void setViewName(String viewName) {
     this.viewName = viewName;
   }
+
+  /**
```
hazendaz, line 84: This code is a combination of two methods from tomcat.  They are both removed.  Applied them together in this case class so it behaves the same as it did.  I've migrated to new FileUpload usage as tomcat did as well but needed this still.  Maybe there is a better way but this was quick and dirty solution.

